PR #2, opened 2023-05-30
Implement log_limit

luizluca:
Without a log_limit, the router with logs enabled in a zone is vulnerable to a DoS attack, with potential traffic amplification if the router sends the log to a syslog server.

With fw3, the router was handling about 60k/h, with peaks of 70k/h. After the upgrade to fw4, the same router started to send reject messages at 320k/h, increasing the routed traffic latency to about 2000ms. I don't know if fw4 is simply generating more logs with the same traffic or if fw3 had an implicit log limit (intentional or not). Maybe fw4 should have an implicit log limit as well.

This series implement the log_limit that was missing in fw4. It creates a dedicated rule only for log action before the other action rules. Also, it adds log_rate to rules (to let the user fine tune log rate for important rules).

It still lacks the test cases but the existing one were updated to deal with the dedicated log rule. I'll add those tests after this PR idea got accepted.

### DIFF
--- a/root/usr/share/firewall4/templates/ruleset.uc
+++ b/root/usr/share/firewall4/templates/ruleset.uc
@@ -2,6 +2,7 @@
 	let flowtable_devices = fw4.resolve_offload_devices();
 	let available_helpers = filter(fw4.helpers(), h => h.available);
 	let defined_ipsets = fw4.ipsets();
+	let zones_with_limits = filter(fw4.zones(), z => z.log_limit);
 -%}
 
 table inet fw4
@@ -82,6 +83,19 @@ table inet fw4 {
 
 {% endfor %}
 
+{% if (length(zones_with_limits)): %}
+	#
+	# Limits
+	#
+
+{%   for (let zone in fw4.zones()): %}
+{%     if (zone.log_limit): %}
+	limit {{ zone.name }}_log_limit { rate {{zone.log_limit.val}} ; comment "{{zone.name}} log limit"; }
+
+{%     endif %}
+{%   endfor %}
+
+{% endif %}
 	#
 	# User includes
 	#
@@ -131,7 +145,7 @@ table inet fw4 {
 		ct state invalid drop comment "!fw4: Drop flows with invalid conntrack state"
 {% endif %}
 {% for (let rule in fw4.rules("forward")): %}
-		{%+ include("rule.uc", { fw4, zone:null, rule }) %}
+		{%+ include("rule.uc", { fw4, zone:(rule.src.zone?.log_limit ? rule.src.zone : rule.dest.zone), rule }) %}
 {% endfor %}
 {% for (let zone in fw4.zones()): for (let rule in zone.match_rules): %}
 		{%+ include("zone-jump.uc", { fw4, zone, rule, direction: "forward" }) %}
@@ -245,6 +259,8 @@ table inet fw4 {
 {%  fw4.includes('chain-append', `forward_${zone.name}`) %}
 		jump {{ zone.forward }}_to_{{ zone.name }}
 {%  if (fw4.forward_policy() != "accept" && (zone.log & 1)): %}
+{%    if (zone.log_limit): %}
+		limit name "{{ zone.name }}_log_limit" {%+ endif -%}
 		log prefix "{{ fw4.forward_policy() }} {{ zone.name }} forward: "
 {%  endif %}
 	}
--- a/root/usr/share/firewall4/templates/ruleset.uc
+++ b/root/usr/share/firewall4/templates/ruleset.uc
@@ -108,7 +108,7 @@ table inet fw4 {
 		tcp flags & (fin | syn | rst | ack) == syn jump syn_flood comment "!fw4: Rate limit TCP syn packets"
 {% endif %}
 {% for (let rule in fw4.rules("input")): %}
-		{%+ include("rule.uc", { fw4, rule }) %}
+		{%+ include("rule.uc", { fw4, zone:null, rule }) %}
 {% endfor %}
 {% for (let zone in fw4.zones()): for (let rule in zone.match_rules): %}
 		{%+ include("zone-jump.uc", { fw4, zone, rule, direction: "input" }) %}
@@ -131,7 +131,7 @@ table inet fw4 {
 		ct state invalid drop comment "!fw4: Drop flows with invalid conntrack state"
 {% endif %}
 {% for (let rule in fw4.rules("forward")): %}
-		{%+ include("rule.uc", { fw4, rule }) %}
+		{%+ include("rule.uc", { fw4, zone:null, rule }) %}
 {% endfor %}
 {% for (let zone in fw4.zones()): for (let rule in zone.match_rules): %}
 		{%+ include("zone-jump.uc", { fw4, zone, rule, direction: "forward" }) %}
@@ -153,7 +153,7 @@ table inet fw4 {
 		ct state invalid drop comment "!fw4: Drop flows with invalid conntrack state"
 {% endif %}
 {% for (let rule in fw4.rules("output")): %}
-		{%+ include("rule.uc", { fw4, rule }) %}
+		{%+ include("rule.uc", { fw4, zone:null, rule }) %}
 {% endfor %}
 {% for (let zone in fw4.zones()): %}
 {%  for (let rule in zone.match_rules): %}
@@ -216,7 +216,7 @@ table inet fw4 {
 	chain input_{{ zone.name }} {
 {%  fw4.includes('chain-prepend', `input_${zone.name}`) %}
 {%  for (let rule in fw4.rules(`input_${zone.name}`)): %}
-		{%+ include("rule.uc", { fw4, rule }) %}
+		{%+ include("rule.uc", { fw4, zone, rule }) %}
 {%  endfor %}
 {%  if (zone.dflags.dnat): %}
 		ct status dnat accept comment "!fw4: Accept port redirections"
@@ -228,7 +228,7 @@ table inet fw4 {
 	chain output_{{ zone.name }} {
 {%  fw4.includes('chain-prepend', `output_${zone.name}`) %}
 {%  for (let rule in fw4.rules(`output_${zone.name}`)): %}
-		{%+ include("rule.uc", { fw4, rule }) %}
+		{%+ include("rule.uc", { fw4, zone, rule }) %}
 {%  endfor %}
 {%  fw4.includes('chain-append', `output_${zone.name}`) %}
 		jump {{ zone.output }}_to_{{ zone.name }}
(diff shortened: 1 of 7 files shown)


root/usr/share/firewall4/templates/ruleset.uc
```diff
@@ -216,7 +222,9 @@ table inet fw4 {
 	chain input_{{ zone.name }} {
 {%  fw4.includes('chain-prepend', `input_${zone.name}`) %}
 {%  for (let rule in fw4.rules(`input_${zone.name}`)): %}
-		{%+ include("rule.uc", { fw4, rule }) %}
+	{%- if (rule.log): %}
+		{%+ include("rule.uc", { fw4, rule, log: true }) %} {%- endif %}
```
brada4, line 237: Why not `log: rule.log` in one line?

luizluca: "log" does not mean "do the action and log" but "do the log action". That's why "log: rule.log" is not enough.

luizluca:
@jow- @nbd168 , could you please take a look?

jow-:
Please restrict the rule duplication to cases where a log_limit is set.

luizluca:
> Please restrict the rule duplication to cases where a log_limit is set and emit combined rules in all other cases.

@jow-, done. I changed how I duplicate rules from using include twice in ruleset to include itself on each template. It looks cleaner but I had to add a double tab before the include to match the indentation. I don't know if there is a better way to do it.

I also added support for log_limit inside redirect entries as I did for rules before. When present, it will overwrite the zone log_limit settings for that rule/redirect. BTW, https://openwrt.org/docs/guide-user/firewall/firewall_configuration does not even mention a rule/redirect can have a log option. That doc needs some update to match fw4.

I passed the zone to rule.uc whenever possible and set it as null when I didn't have a zone (i.e.: input rules without a zone). For forward rules, I'm using, in order, src or dest zone that has log_limit defined. I'm not sure if it is the right way to implement it.

I didn't fix the ucode indentation after I added an if around the code. It reduced the amount of changed code, making the review much cleaner. I can fix that if required.

I'm adding the "# Limits" header unconditionally as others like "User includes" do. Should I check if any zone sets log_limit and avoid adding it when not required?

I didn't limit MSSFIX log. I believe it decides whether to log after it fixes the mss. I would need to detect beforehand if mssfix would do anything to log (with limit) in a dedicated clause.

I added proper tests for the added features. It will still fails because of #5

tests/05_ipsets/02_usage
```diff
@@ -146,6 +146,10 @@ table inet fw4 {
 	#
 
 
+	#
+	# Limits
```
brada4, line 150: Do you have a coverage metric to fill by these nop tssts?

luizluca: I was adding that header unconditionally but reviewing the code, fw4 avoids adding headers when not needed. I just pushed a fix adding Limits section only when needed.

root/usr/share/firewall4/templates/ruleset.uc
luizluca: I changed this logic quite a bit. I'll set this as resolved. Now the rule duplication happens inside each template.

luizluca:
New push with "# Limits" header only appearing when there is a zone with log_limit. If fw4 adds support for other limits, that same section might be expanded for those cases too.

brada4:
Thank you, let me suggest other test - with logs disabled no log statements land in emitted ruleset (it passes ;-) )

luizluca:
> Thank you, let me suggest other test - with logs disabled no log statements land in emitted ruleset (it passes ;-) )

I added a new guest zone with no log but log_limit and two rules, one with log (that will use log_limit).

jow-:
I wonder if we should utilize [anonymous chains](https://lwn.net/Articles/824594/) for this use case, which would eliminate my biggest pet peeve, the need to match packets twice, once for the log rule and once for the actual target.

brada4:
You can "log" basically in every whitespace even between chained matches or in lone otherwise blank line, like counter.

luizluca:
> I wonder if we should utilize [anonymous chains](https://lwn.net/Articles/824594/) for this use case, which would eliminate my biggest pet peeve, the need to match packets twice, once for the log rule and once for the actual target.

That's new for me. Thanks. It isn't mentioned in docs (https://wiki.nftables.org/). It is indeed a nice feature.

Instead of self including the rule/redirect, we could prefix the action part of the rule with "chain {\n" + "\t\t\t" + \<log rule\> + "\n" + "\t\t\t" and add an extra closing "}" after the action. I can adapt this PR that way if requested.

However, if firewall4 wants to use anonymous chains in the future for any other case, it would be better to have some internal support for subrules and not simply inject it in the template for log_limit. Are there any plan for that? The UCI syntax might limit a little how we could explore subrules. I believe most cases supported by current UCI syntax might as well be optimized simply using anonymous sets, maps, etc.

I wonder if a jump+ret would cost much more than a double check. As a compassion, firewalld shamelessly repeats checks:

```
                iifname "enp0s31f6" oifname "docker0" jump filter_FWD_public
                iifname "enp0s31f6" oifname "docker0" meta pkttype host log prefix "filter_FWD_public_REJECT: "
                iifname "enp0s31f6" oifname "docker0" reject with icmpx admin-prohibited
```

We could also merge this as is and improve/optimize rule generation in the future.

luizluca:
I debugged #5. I know why it is generating bogus jumps but I don't what is the right fix. I can, if needed, workaround in this PR adding target to all failing rules.

tests/02_zones/08_log_limit
```diff
@@ -0,0 +1,438 @@
...
+	}
+
+	chain handle_reject {
+		meta l4proto tcp reject with tcp reset comment "!fw4: Reject TCP traffic"
+		reject with icmpx type port-unreachable comment "!fw4: Reject any other traffic"
+	}
+
+	chain input_lan {
+		tcp dport 1 limit name "lan_log_limit" log prefix "@rule[0]: "
+		tcp dport 1 counter comment "!fw4: @rule[0]"
```
brada4, line 272: https://en.m.wikipedia.org/wiki/TCP_Port_Service_Multiplexer

Please avoid super dangerous ports in examples

brada4:
You can ratelimit in rsyslog/syslog-ng already.

jow-:
> Instead of self including the rule/redirect, we could prefix the action part of the rule with "chain {\n" + "\t\t\t" + <log rule> + "\n" + "\t\t\t" and add an extra closing "}" after the action. I can adapt this PR that way if requested.

I'd appreciate if you could give this approach a try, it would feel cleaner to me.

> However, if firewall4 wants to use anonymous chains in the future for any other case, it would be better to have some internal support for subrules and not simply inject it in the template for log_limit. Are there any plan for that?

I don't have any plan to expose this functionality in uci directly, but I consider using it as internal implementation detail to improve the ruleset in some cases. E.g. for uci rules using both icmp and tcp/udp protocols, we could combine the common match bits in a single rule and handle the differences such as icmp type vs. sport/dport in an anonymous chain.

brada4:
The double-rule logging for good states is not guaranteed to log objectively nearing full conntrack table.
I would suggest to have 2-rule log at the mini-chains that do final reject/drop without additional filter akin firewalld, not mandating duplicate inspection filters.
State activities can be logged by (not packaged) conntrackd in addition or place  to its main call of syncing HA firewalls.
Once integrated that is lightweight with clear improvement path.

luizluca:
> You can ratelimit in rsyslog/syslog-ng already.

The issue is not to avoid simply avoid a lot of logs but that a lot of calls to nft log (in my test 320k/h or 80/s) starts to affect the network traffic, increasing the latency to over 2000ms. We cannot limit that afterwards.

tests/02_zones/08_log_limit
luizluca: I will all 1000 to all ports.

luizluca:
> > Instead of self including the rule/redirect, we could prefix the action part of the rule with "chain {\n" + "\t\t\t" +  + "\n" + "\t\t\t" and add an extra closing "}" after the action. I can adapt this PR that way if requested.
> 
> I'd appreciate if you could give this approach a try, it would feel cleaner to me.

I did just to fail when I tried it in a real system (fresh snapshot image):

```
root@OpenWrt:/# grep RELEASE /etc/os-release 
OPENWRT_RELEASE="OpenWrt SNAPSHOT r23727-e6f8b69918"
root@OpenWrt:/# cat a.nft 
 table inet x {
        chain y {
                type filter hook input priority 0;
                tcp dport 22 chain {
                        ip saddr { 127.0.0.0/8, 172.23.0.0/16, 192.168.13.0/24 } accept
                        ip6 saddr ::1/128 accept;
                }
        }
 }
root@OpenWrt:/# nft -f a.nft 
a.nft:4:30-34: Error: syntax error, unexpected chain, expecting newline or semicolon
                tcp dport 22 chain {
                             ^^^^^
a.nft:9:2-2: Error: syntax error, unexpected '}'
 }
 ^
```

The a.nft is from the commit message you mentioned. Is userland ready? I also tried in a full Linux system (OpenSUSE Tumbleweed / kernel 6.4.8-1-default) with the same results.

> I do have another general remark wrt. limits; I believe we should not name the limits after the zones using them but after their value, so that log rules or log zones using the same limit values can share the named limits. So it would not be `wan_log_limit` but something like `log_limit_100ps` (where `100` would be `x.log_limit.rate` and `s` would be `substr(x.log_limit.unit, 0, 1)`).

I'm not sure about this. A declared limit is not just a "macro". In my tests, it looks like it shares the same counter with all rules that uses it. If zone_a.rule_a used all the available slots for zone_a, zone_a.rule_b will be silent until the log slots are refreshed. If two zones uses log_limit='10/minute', using a common "log_limit_10pm" will allow only 10 logs per minute from both zones, not for each zones. We can rename the limit name used by log_limit to whatever fits better but we cannot implicitly share it with other rules/zones that are coincidentally using the same rate. I still thing the current name is fine.

luizluca:
> The double-rule logging for good states is not guaranteed to log objectively nearing full conntrack table.
> I would suggest to have 2-rule log at the mini-chains that do final reject/drop without additional filter akin firewalld, not mandating duplicate inspection filters.
> State activities can be logged by (not packaged) conntrackd in addition or place to its main call of syncing HA firewalls.
> Once integrated that is lightweight with clear improvement path.

As the anonymous chain does not seem to be available, the only way I see to implement this without double matching is to use declared chains like zoneA_log_n_{accept,reject,drop,...}, with extra chains whenever the rule uses a custom log_rate ruleX_log_n_(accept,reject,drop,...). It would be a lot of new chains and a lot of jumps. I'm not sure it is worth it as it will pollute the ruleset and probably costs even more. What I do know is that today, any router with fw4 and log enable is vulnerable to a DoS attack if the attacker knows how to trigger a log. We do need something to protect the router, even if we miss some matches/logs in some situations.

luizluca:
I just pushed a rebased branch with ports increased to 1000+x. It is still based on double matching. With the new fixes @jow- did in master, all tests are OK.

jow-:
Merged via a923c88f9fbb055247cf0a0903f4ad1c189a4843, 597dc90fb71419c42c13c27fe24e3d350e6e33dc, 187405075911d408fa48e97ce343e76a2a30ef12 with some slight refactoring.

The most serious problem was the formatting of the rate expressions. We must not use `{{ rate.val }}` (which is the original config string) but the normalized `{{ rate.rate }}` and `{{ rate.unit }}` values. This is needed because the uci config accepts abbreviations like `4` or `4/s` while nft expects `4/second`.

PR #3, opened 2023-06-21
Apply MSS fixup in place with final MTU before wire.

brada4:
- locate mss fixup in postrouting table so that final route MTU is used in nat/docker forwarding and mwan scenario where incoming, most likely 1500 was applied
-  reduce scope of fixup to only legit SYN and SYN+ACK packets potentially pertaining connection establishment
-  one final rule 2x faster than existing 2

Fixes: https://github.com/openwrt/openwrt/issues/12112
Part-fixes: https://github.com/openwrt/openwrt/issues/12805

### DIFF
--- a/root/usr/share/firewall4/templates/ruleset.uc
+++ b/root/usr/share/firewall4/templates/ruleset.uc
@@ -417,6 +417,13 @@ table inet fw4 {
 {% for (let rule in fw4.rules("mangle_postrouting")): %}
 		{%+ include("rule.uc", { fw4, rule }) %}
 {% endfor %}
+{% for (let zone in fw4.zones()): %}
+{%  if (zone.mtu_fix): %}
+{%   for (let rule in zone.match_rules): %}
+                {%+ include("zone-mssfix.uc", { fw4, zone, rule, egress: true }) %}
+{%   endfor %}
+{%  endif %}
+{% endfor %}
 {% fw4.includes('chain-append', 'mangle_postrouting') %}
 	}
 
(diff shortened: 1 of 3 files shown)


brada4:
Fixes: https://github.com/openwrt/openwrt/issues/13036

brada4:
@jow- could you consider fixing this long standing logic bug?

jow-:
I am reluctant to merge this, as I do not fully understand the change. I can find no examples on the internet placing "set rt mtu" rules in postrouting, only ever in filter/forward.

brada4:
In filter/forward  state egress MTU is not known as per https://github.com/openwrt/openwrt/issues/12112#issuecomment-1454671477
Actually incoming interface MTU gets applied https://github.com/openwrt/openwrt/issues/12805 (steps to reproduce)

brada4:
mtu fixup as per `man 8 iptables-extensions` only works in `mangle` table, any of 5 chains 
```
*mangle
:PREROUTING ACCEPT [0:0]
:INPUT ACCEPT [0:0]
:FORWARD ACCEPT [0:0]
:OUTPUT ACCEPT [0:0]
:POSTROUTING ACCEPT [0:0]
COMMIT
```

https://www.digitalocean.com/community/tutorials/a-deep-dive-into-iptables-and-netfilter-architecture#chain-traversal-order

brada4:
From around internet - shorewall(iptables) and firewalld(nftables) both use one global MSS clamp, which solves the enterprise challenge of having jumbo frames mixed with normal ones, then connecting multiple such locations via (Linux) router. i.e whichever interface receives jumbo-sized SYN MSS it gets fixed for "our" MTU. 
Here we have much wilder MTU variation thus we need more detailed approach.

Whats interesting - firewalld has mss clamping in filter/forward which likely means filter/postrouting could be a valid location, but absence of that table leaves us with lower level mangle/postrouting only.

brada4:
@jow-  You can leave input mssfix if you want, but this is essential to bring output at lower level. the "other" rule is is general input mssfix on forward table which saves mtu9000 machines communicating over mtu-challenged routers. Like you connect machines with varied mtu to same network and it gets fixed just by having firewall enabled. It is a bit different than encap-encap-encap internet connections.

jow-:
Merged via 698a53354fd280aae097efe08803c0c9a10c14c2 - thanks!

brada4:
I tjink ../openwrt 10224 will be silently fixed too.... Lets hope someone confirms in the meantime.

PR #4, opened 2023-06-23
Disable loopback state tracking to match firewall3

brada4:
- disable conntrack-ing loopback connections akin iptables-legacy https://github.com/openwrt/openwrt/issues/12522

old iptables treated all rules before first conntrack rule as stateless, with nftables this became unnecessary flexing of conntrack before it actually does anything useful.

- enact positive state machine first before per-packet examination

While pre-bulk per-packet processing is drastically reduced, still accept conntrack matches first as done with conntrack offloads. 

Fixes: https://github.com/openwrt/openwrt/issues/12522

### DIFF
--- a/root/usr/share/firewall4/templates/ruleset.uc
+++ b/root/usr/share/firewall4/templates/ruleset.uc
@@ -99,7 +99,7 @@ table inet fw4 {
 
 {% fw4.includes('chain-prepend', 'input') %}
 		ct state established,related accept comment "!fw4: Allow inbound established and related flows"
-		iifname "lo" notrack accept comment "!fw4: Accept traffic from loopback"
+		iifname "lo" accept comment "!fw4: Accept traffic from loopback"
 {% if (fw4.default_option("drop_invalid")): %}
 		ct state invalid drop comment "!fw4: Drop flows with invalid conntrack state"
 {% endif %}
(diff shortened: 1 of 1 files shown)


brada4:
@jow- hi, notrack needs raw option for nftables in kernel, re-ordering alone gives shortcut already.

brada4:
"new" ct notrack works in lower priority tables only unlike old notrack.
this reduces network jitter as measured by ntpd behind NAT from hundredths to sub-thousandths of second.

jow-:
So the idea of this change is to move the iifname/oifname lo handling after the conntrack related match to prevent matching *every* packet against lo?

brada4:
Yes, and npt create never used states for loopback

brada4:
iptables-legacy filter  rules before any use of conntrack (incl reject) were stateless, this best/smallest step closer to greatness.

brada4:
What is happening now - 'lo' states get created, but continuation gets accepted with status 'none' outside those states. Which is unnecessary effort for 'lo' states, and unwillingly for primary function - NAT from LAN to WAN.

brada4:
@jow- found correct place to stop tracking always allowd anyway loppback connections.

jow-:
Not sure if it is a good idea to make this default and not even providing an option to turn it off. Existing deployments might rely on loopback traffic being conntracked.

brada4:
It is not conntracked, it is accepted as new, or as invalid with undue delay.

brada4:
Last user gives up on 127/8 https://github.com/kubernetes/kubernetes/issues/90259

brada4:
Sort of not critical to primary functions.

PR #6, opened 2023-08-26
Fix excess responses on noisy WAN network 

brada4:
Fixes: https://github.com/openwrt/openwrt/issues/13340

In principle there is nothing forwarded from WAN unless some forward rule is present or it tries NAT traversal on recent new connection attempt. Past fixing excessive load generated in reported case also nat traversal will not need to retry socket connection, retransmitted SYN is likely to come when permitting rule/transient state is completely established.

### DIFF
--- a/root/etc/config/firewall
+++ b/root/etc/config/firewall
@@ -2,7 +2,7 @@ config defaults
 	option syn_flood	1
 	option input		REJECT
 	option output		ACCEPT
-	option forward		REJECT
+	option forward		DROP
 # Uncomment this line to disable ipv6 rules
 #	option disable_ipv6	1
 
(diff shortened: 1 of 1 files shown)


jow-:
I could live with defaulting WAN-to-WAN forward to drop. The global inter-zone forward policy should stay at reject though.

brada4:
OK, the global policy applies on new interfaces, but im fine with endangering unsuspecting.

brada4:
Test case: connect wan to (10mbps) hub/repeater and watch spoofed traffic killing all communication there.

brada4:
Sorry, wrong commit headings.

PR #7, opened 2023-08-31
Remove duplicate interface list match in verdict chain

brada4:
Interface name list per zone is already filtered in basic chains then via jumps the arriving traffic is only that already filtered by interfaces.
Just dont emit 2nd {nterface list} @jow- 


jow-:
NAK, this will break for example the following rule:

```
config rule
 option src *
 option dest wan
 option proto 42
 option target DROP
```

Due to the removal of the egress interface matches, the above rule would drop *any* forwarded protocol 42 traffic, not just traffic directed at one of the wan zone devices.

Also the jumps leading up the `$action_to_$zone` chains match the ingress devices (`iif` / `iifname`) while the matches you removed match the egress ones (`oif` / `oifname`).

brada4:
OK, will rework to generate iface filter in upper chain rules

PR #8, opened 2023-09-01
rate-limit ICMP errors similar to kernel defaults

brada4:
Rationale described here:
https://github.com/openwrt/openwrt/issues/13340#issuecomment-1693315898

UDP closed port will return 80 octets for 64 octets input at 1000/sec for each ip6/ip6 by kernel while nftable will have no limit. 
It is quick as in efficient approximation of more sophisticated linux kernel behaviour. Namely it does not implement 1 packet per address per timeslice on selected ICMP types.

### DIFF
--- a/root/usr/share/firewall4/templates/ruleset.uc
+++ b/root/usr/share/firewall4/templates/ruleset.uc
@@ -194,6 +194,7 @@ table inet fw4 {
 				? `icmpx type ${fw4.default_option("tcp_reject_code")}`
 				: "tcp reset"
 		}} comment "!fw4: Reject TCP traffic"
+		limit rate over 1000/second burst 50 packets counter drop
 		reject with {{
 			(fw4.default_option("any_reject_code") != "tcp-reset")
 				? `icmpx type ${fw4.default_option("any_reject_code")}`


brada4:
ping @jow-

jow-:
The limit needs to be configurable, we shouldn't add hardcoded limits to the ruleset.

brada4:
yup, this exceeds default unconfigurable syn flood  limit by magnitude.

jow-:
> yup, this exceeds default unconfigurable syn flood limit by magnitude.

??

brada4:
Syn limit is 25/s enable/disable.
Could we agree to set this to 1000/s and add option to reduce it if anyone ever asks? Sort of all commercial routers limit generated ICMP unavoidably with 1000/s being configurable maximum with defaults anywhere below.
At present it generates a bit more responses (84B in 100B out measured on ethernet) than ping, so it can even bring down upstream of symmetrical link. 1.2x amplification is not that much of concern globally.

jow-:
> Syn limit is 25/s enable/disable.

It *defaults* to 25/s and can be adjusted using the `synflood_rate` and `synflood_burst` options. Your change would need to introduce new `reject_rate` and `reject_burst` options  which may default to `1000/sec` and `50` respectively.

I just don't want to introduce any kind of *hardcoded* limits into the ruleset. Also please adjust the testcases accordingly.

brada4:
On weekend more.

brada4:
@jow- is there a resource in ucode that takes sysctl values or it is to exec them?

jow-:
If there's a corresponding `/sys/` or `/proc/sys/` file then just `readfile()` it, otherwise `popen()` + `read()`

PR #9, opened 2023-09-02
Make conntrack checks in single rule

brada4:
vmap of 2 states is faster than state,state, in addition handle 3rd state in same CPU time
reduces 10ms->9ms->7ms idle latency as measured at waveform on TL-WR1043NDv1 v22.03
Very steady measured at different day times.
@jow- ?

credits: `nft -o`


jow-:
Please use consistent white space and adjust the testcases accordingly.

brada4:
see other

PR #10, opened 2023-09-07
dispatch states using verdict map

brada4:
Obey optimisations suggested by nft -o
Makes new connections less demanding, measurably so on sub-par CPU-s
BTW whitespace after tag gets wiped off.

Signed-Off-By: `Andris PE <neandris..gmail.com>`

PR #11, opened 2023-09-19
Halve ksoftirq load by refering to loopback interface by fixed index 1 in place of matching text.

brada4:
Reduce ksoftirq load by half using more efficient reference to loopback which always has index equal to one.
Should help a lot with https://github.com/openwrt/openwrt/issues/12914 https://github.com/openwrt/openwrt/issues/12121 and similar.
iperf3 is single-core, so on single processor it is pressure vs ksoftirq from firewall, on multicore less load on one core gives more hertz to other.

@jow- backports to v22 directly

### DIFF
--- a/root/usr/share/firewall4/templates/ruleset.uc
+++ b/root/usr/share/firewall4/templates/ruleset.uc
@@ -97,7 +97,7 @@ table inet fw4 {
 	chain input {
 		type filter hook input priority filter; policy {{ fw4.input_policy(true) }};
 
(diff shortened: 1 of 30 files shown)


jow-:
Change looks fine.

- Please update the commit message to contain this PR description
- Add a proper `Signed-off-by` to your commit

brada4:
Please also check #10  for nearby mandatory element comparing strings.

jow-:
Merged via a5553dae70439c7e4fa910490fcf12a1ffff5bd2 - thanks!

PR #20, opened 2024-01-18
Reorder offload processing flow for earlier pickup

brada4:
Do ct state dispatch before offload attempt on established states only.

Before:
- first data packet in connection commonly TLS client hello is attempted to be offloaded without assured ct status then follows slow path and response , normally tls server hello , offloads flow and follows the slow path again .
- It takes 2 slow paths to re-offload timer-retired or window-adjusted packets to re-enter offload table . 

After:
- ct status is marked as assured and packet is offloaded and response does not traverse slow path .
- Retired flow is adjusted by single packet and subsequent exchange offloaded skipping attempt to stuff them back in unmodified ct+offload table .
- ICMP (frag needed for example) reach related flow and offload happens on next TCP/UDP packet as before (minus the unnecessary match against udp in present path) .
- Less(er) minutely hickups in games and improved UDP VPN latency .

### DIFF
--- a/root/usr/share/firewall4/templates/ruleset.uc
+++ b/root/usr/share/firewall4/templates/ruleset.uc
@@ -134,11 +134,9 @@ table inet fw4 {
 	chain forward {
 		type filter hook forward priority filter; policy {{ fw4.forward_policy(true) }};
 
-{% if (length(flowtable_devices) > 0): %}
-		meta l4proto { tcp, udp } flow offload @ft;
-{% endif %}
 {% fw4.includes('chain-prepend', 'forward') %}
-		ct state vmap { established : accept, related : accept{% if (fw4.default_option("drop_invalid")): %}, invalid : drop{% endif %} } comment "!fw4: Handle forwarded flows"
+		ct state vmap { established : {% if (length(flowtable_devices) > 0): %} jump handle_offload {% else %} accept {% endif -%}
+			, related : accept {% if (fw4.default_option("drop_invalid")): %}, invalid : drop{% endif %} } comment "!fw4: Handle forwarded flows"
 {% for (let rule in fw4.rules("forward")): %}
 		{%+ include("rule.uc", { fw4, zone: (rule.src?.zone?.log_limit ? rule.src.zone : rule.dest?.zone), rule }) %}
 {% endfor %}
@@ -207,6 +205,13 @@ table inet fw4 {
 		}} comment "!fw4: Reject any other traffic"
 	}
 
+{% if (length(flowtable_devices) > 0): %}
+	chain handle_offload {
+		meta l4proto { tcp, udp } flow offload @ft
+		accept
+	}
+
+{% endif %}
 {% if (fw4.default_option("synflood_protect") && fw4.default_option("synflood_rate")):
 	let r = fw4.default_option("synflood_rate");
 	let b = fw4.default_option("synflood_burst");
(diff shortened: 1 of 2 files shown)


brada4:
@jow- PING, thanks.

CallMeR:
Perhaps including flows with an established or related connection state in the offload could be a better choice :)

```bash
chain forward {
        ct state established,related flow add @ft 
        ct state vmap { established : accept, related : accept, invalid : drop }
    }
```

brada4:
at present offload engine does not promote !(tcp udp) packets present in related (as opposed to helper expected packets which will be offloaded once they pass from NEW to established) at all, if scope changes this can be changed.

brada4:
@CallMeR but you have an interesting point. 
Is omitting tcp,udp meta faster to skip offload than offloads own filter?

I do one state vmap dispatch, remember packet that initiates offload leaves nft only after slow path is complete. You can drop tcp one to trigger retransmit.
In your example estb packet state will stll hit 2nd accept, same flow.

Some interesting examples:
```
counter
flow add counter # drop
counter
```
```
add offloads in a separate related branch
none offloads...
```

brada4:
[i mean have no meta here](https://github.com/openwrt/firewall4/pull/20/files#diff-67bacc8013b62bb45df880a71296fb7ea3eb467ed246a71f073a3d323d76b2d5R210)

brada4:
In my tests fastest is
```
flow add @ft accept
accept
```
apachebench lan against nginx wan with 1000 connections for 1m total 1043v2 (old CPU-challenged router)
timer resolution seems 1ms and its about 1/3ms average adding extra accept and 1/20ms average omitting meta filter.

brada4:
@CallMeR send mail with name if you want credit for idea in description.

CallMeR:
> @CallMeR send mail with name if you want credit for idea in description.

Thank you!  It is already enough for me if it can help make fw4 better :)
Please feel free to submit this idea on your behalf.

brada4:
Did it alr, thanks for insightful feedback.

brada4:
Just being crybaby - no dnat in luci.

brada4:
@jow-  could you please attend this?
- locate offload yielding to kwork at the end of slowpath
- drop invalid asap which accidentally simplifies main state dispatch
Still:
- do we need loopback nmap pedantically support
- there is an opportunity to eg counter+drop lo forward based on sysctl value ie catch numbers before sysctl drops packets.

jow-:
Didn't you just recently add the vmap dispatching? Now it is partially rolled back? Also please squash the back-and-forth-commits in this PR into a single commit and try to explain exactly what is improved and how.

brada4:
vmap(a:x,b:x) is identical  while harder to read than a,b x in representing a x;b x
vmap(a:x,b:y) is quicker form of a x;b y
Just that it gave nice text processing shortcut for 3 identical places.
see now it is still present in pre-rendered places where verdict outcomes are different. 

if there was a utility function simulating vmap that would collapse to comma list in cases of single outcome, that could be used say in place of lan->jump lan, wan->jump wan, iot-> jump iot, but for now it is pre-rendered vmap-s where multiple rules are otherwise unavoidable.

Changed gitconfig, sorry for that. Will make clean PR instead. You can squash this as you read it.

brada4:
iifname vmap { "virbr0" : accept } is as fast as iif "virbr0" accept, but catches interface id change, go figure.

PR #23, opened 2024-02-29
Drop to-be-forwarded-nowhere packets on wans

brada4:
Dropping packets with no clear forward destination is nicer than rejecting them. Especially when some providers punish users for spoofing caused by their noisy infra.

Fixes: https://github.com/openwrt/openwrt/issues/13340

### DIFF
--- a/root/etc/config/firewall
+++ b/root/etc/config/firewall
@@ -19,7 +19,7 @@ config zone
 	list   network		'wan6'
 	option input		REJECT
 	option output		ACCEPT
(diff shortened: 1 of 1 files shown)


jow-:
Merged via 97962771aa3c490d6186e64015f85dd66254fdf0 with slightly changed commit subject, rewrapped commit message and S-o-b added on your behalf (taken from the PR description).

brada4:
I think i understtood how sob is lost, ill rty next better

PR #25, opened 2024-03-12
Do not resolve physical devices for soft offload

brada4:
Let kernel heuristics take care of offloading decapsulation Packets may still enter flow engine one encapsulation below actual interface subject to heuristics, while exiting it on listed interfaces, in kernel subject to non-flow encapsulation offloads.

Fixes: https://github.com/openwrt/openwrt/issues/13410
Fixes: https://github.com/openwrt/openwrt/issues/10224

This is minimally intrusive, to address issue, likely should be bitmask to selectively enable (forward interfaces|one encap below|virt interfaces|wifi|wifi not unique AP|probably others) or flaggs to +/- exact interface names to list.

### DIFF
--- a/root/usr/share/ucode/fw4.uc
+++ b/root/usr/share/ucode/fw4.uc
@@ -520,7 +520,11 @@ return {
 
 		for (let zone in this.zones())
 			for (let device in zone.related_physdevs)
-				push(devices, ...resolve_lower_devices(devstatus, device));
+				if (this.default_option("flow_offloading_hw")) {
+					push(devices, ...resolve_lower_devices(devstatus, device));
+				} else {
+					push(devices,device);
+				}
 		devices = sort(uniq(devices));
 
 		if (this.default_option("flow_offloading_hw")) {
--- a/root/usr/share/ucode/fw4.uc
+++ b/root/usr/share/ucode/fw4.uc
@@ -2049,9 +2049,6 @@ return {
 					});
 				}
 
-				if (net.physdev && !e.invert)
-					push(related_physdevs, net.physdev);
-
 				push(related_subnets, ...(net.ipaddrs || []));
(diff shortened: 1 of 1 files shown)


brada4:
@jow- ? Or it would need to check if l3dev then l3dev?
```diff
--- a/root/usr/share/ucode/fw4.uc
+++ b/root/usr/share/ucode/fw4.uc
@@ -599,7 +599,7 @@ return {
                        for (let ifc in ifaces.interface) {
                                let net = {
                                        up: ifc.up,
-                                       device: ifc.l3_device ?? ifc.device,
+                                       device: ifc.l3_device,
                                        physdev: ifc.device,
                                        zone: ifc.data?.zone
                                };
@@ -647,7 +647,7 @@ return {
                                                        ...rulespec,
 
                                                        name: (rulespec.type != 'ipset') ? `ubus:${ifc.interface}[${ifc.proto}] ${rulespec.type || 'rule'} ${n}` : rulespec.name,
-                                                       device: rulespec.device ?? ifc.l3_device ?? ifc.device
+                                                       device: rulespec.device ?? ifc.l3_device
                                                });
 
                                                n++;
```

brada4:
@jow- ok, side effect is  issue 10224 is partially addressed, not completely fixed. Should I polish message to less happy?

brada4:
For the other part of issue is TBD, seems netifd sometimes (pppoe) adds l3 device which is kind of to transfer all rules to, sometimes it has only l3device while GRE tunnel has no routes etc. No idea as of yet regarding needed logic vs netifd emissions

brada4:
Ill leave it half-way. Fixes: roaming Fixes half pppoe, nothing more.

brada4:
will repeat with clean tree.

PR #27, opened 2024-03-22
ruleset: do not emit helper jump without helper modules

brada4:
Do not emit helper jump if no helper modules are present saving per-every-packet bytecode before ct:
```
inet fw4 prerouting
  [ meta load iifname => reg 1 ]
  [ cmp eq reg 1 0x6c2d7262 0x00006e61 0x00000000 0x00000000 ]
  [ immediate reg 0 jump -> helper_lan ]
  userdata = { \x00-!fw4: Handle lan IPv4/IPv6 helper assignment\x00 }
```


systemcrash:
ping @jow-

brada4:
That iifname takes 1/5-1/4 CPU time just forwarding in default config..... Seems dev tests have more weight than generic default ux.

PR #28, opened 2024-03-28
Auto-accept dstnat mangled flows

brada4:
Magically accept packets and subsequent DNAT flows akin other NAT translations. dstnat are mangled before filter hook and they reach filter/forward in state new, with distinct status dnat and otherwise would need secondary filter/forward accept rule to proceed anywhere.

### DIFF
--- a/root/usr/share/firewall4/templates/ruleset.uc
+++ b/root/usr/share/firewall4/templates/ruleset.uc
@@ -139,6 +139,7 @@ table inet fw4 {
 {% endif %}
 {% fw4.includes('chain-prepend', 'forward') %}
 		ct state vmap { established : accept, related : accept{% if (fw4.default_option("drop_invalid")): %}, invalid : drop{% endif %} } comment "!fw4: Handle forwarded flows"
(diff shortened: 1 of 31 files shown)


jow-:
This seems redundant to the already existing per-zone "Accept port forwards" and "Accept port redirection" rules, which are also only emitted if the related zones uses any DNAT rule.

brada4:
I was thinking vice versa, to terminate evaluation asap.

brada4:
The main problem is anomalously slow ?ifname that on one side delays our packet on other consumes processing where others could run. Maybe a bettwer idea to wrap rule(s) in loop to check if any dnat rule is present then emit central rule?

brada4:
Principal problem is that nat-pmp UDP creates UDP DNAT rule but backs it with TCP ACCEPT and whatsapp glitches more than without any traversal. This is not a solution, but would playnly rule out that problem.

brada4:
I will re-work it to add early accept in presence of any dnat rule eliminating duplicates.

PR #33, opened 2024-06-06
[WIP]Anchor script regex for precise match

brada4:
Anchor grep regex to avoid wrongly matching unrelated inet fw4whatever style table names


brada4:
Got better idea along lines of
nft <<EOF
add table
flush table
EOF

Will PR soon

PR #35, opened 2024-07-27
fw4: when ipset matches MAC, allow set family to 'any'

### LINKED ISSUE #16: fw4 assumes ether_addr sets are ipv4 only
Given the following uci configuration:

```
	add firewall rule
	set firewall.@rule[-1].name='Forward-auth-captive'
	set firewall.@rule[-1].src="captivezone"
	set firewall.@rule[-1].dest='wan'
	set firewall.@rule[-1].proto='any'
	set firewall.@rule[-1].target='ACCEPT'
	set firewall.@rule[-1].ipset='captive'
 
	add firewall ipset
	set firewall.@ipset[-1].name='captive'
	add_list firewall.@ipset[-1].match='src_mac'
 ```

firewall4 produces the following rules:
```
table inet fw4 {
	set freewave {
		type ether_addr
		elements = { }
        }
```
[...]
```
	chain forward_captivezone {
		meta nfproto ipv4 ether saddr @captive counter packets 111598 bytes 30867442 jump accept_to_wan comment "!fw4: Forward-auth-captive"
```

The `meta nfproto ipv4` match is unwarranted as the relative set is address agnostic.

### PR DESCRIPTION
**Pull request by zsien**
When filtering by MAC address, it is usually necessary to filter both IPv4 and IPv6.
If it is not allowed to set the family of ipset to any, it will be necessary to create a separate, identical ipset for both IPv4 and IPv6.

Fixes: https://github.com/openwrt/firewall4/issues/16

### DIFF
--- a/root/usr/share/ucode/fw4.uc
+++ b/root/usr/share/ucode/fw4.uc
@@ -2571,7 +2571,7 @@ return {
 
 			/* check if there's no AF specific bits, in this case we can do an AF agnostic rule */
 			if (!family && rule.target != "dscp" && !has_ipv4_specifics && !has_ipv6_specifics) {
-				add_rule(0, proto, [], [], sports, dports, null, null, null, rule);
+				add_rule(0, proto, [], [], sports, dports, null, null, ipset, rule);
 			}
 
 			/* we need to emit one or two AF specific rules */
@@ -3305,11 +3305,7 @@ return {
 			return;
 		}
 
-		if (ipset.family == 0) {
-			this.warn_section(data, "must not specify family 'any'");
-			return;
-		}
-		else if (!length(ipset.match)) {
+		if (!length(ipset.match)) {
 			this.warn_section(data, "has no datatypes assigned");
 			return;
 		}
@@ -3317,6 +3313,11 @@ return {
 		let dirs = map(ipset.match, m => m[0]),
 		    types = map(ipset.match, m => m[1]),
 		    interval = false;
+
+		if (("ip" in types || "net" in types) && ipset.family == 0) {
+			this.warn_section(data, "must not specify family 'any' when matching type 'ip' or 'net'");
+			return;
+		}
 
 		if ("set" in types) {
 			this.warn_section(data, "match type 'set' is not supported");

### CODE REVIEW

**Review by brada4 — APPROVED** [review]
Does what it says.

**Comment by f00b4r0** [comment]
So I just stumbled upon this PR while looking again at #16 and after testing it I don't see how it "fixes" anything.

The `meta nfproto ipv4` match is still emitted by fw4 for every rule that refers to the set. Sure you can add an `option family 'any'` to the ipset configuration, it will not emit a warning but it changes nothing.

**Comment by zsien** [comment]
> So I just stumbled upon this PR while looking again at #16* and after testing it I don't see how it "fixes" anything.
> 
> The `meta nfproto ipv4` match is still emitted by fw4 for every rule that refers to the set. Sure you can add an `option family 'any'` to the ipset configuration, it will not emit a warning but it changes nothing.

@f00b4r0 To maintain compatibility with the configuration, the default value for the ipset family is still ipv4.

I tested the configuration from #16 as follows:
```
config rule
  option name 'Forward-auth-captive'
  option src 'lan'
  option dest 'wan'
  option proto 'any'
  option target 'ACCEPT'
  option ipset 'captive'

config ipset
  option name 'captive'
  option family 'any'
  list match 'src_mac'
```

The generated nftables rules are:
```
set captive {
  type ether_addr
}
...
chain forward_lan {
  ...
  ether saddr @captive counter packets 0 bytes 0 jump accept_to_wan comment "!fw4: Forward-auth-captive"
  ...
}
```

It can be seen that this modification is effective.

**Comment by brada4** [comment]
I know, it duly prints warning when that is employed.

**Comment by f00b4r0** [comment]
@zsien unfortunately while the test you performed works, this one doesn't:

```
config redirect
        option name 'Redirect-unauth-captive'
        option src 'lan'
        option src_dport '80'
        option proto 'tcp'    
        option target 'DNAT'   
        option reflection '0'
        option ipset '!captive'

config ipset
        option name 'captive'
        option family 'any'
        list match 'src_mac'
```

It fails to load the firewall with:

```
/dev/stdin:236:35-36: Error: syntax error, unexpected !=, expecting newline or semicolon
		meta nfproto ipv4 tcp dport 80  != @captive counter redirect to 80 comment "!fw4: Redirect-unauth-captive"
		                                ^^
```

`meta nftproto ipv4` is still emitted and `ether saddr` is missing.
It's likely that other corner cases are thus broken by your change.

I also don't think that `src_mac` ipsets should still default to ipv4, but that's left to the maintainer's appreciation.

HTH

f00b4r0:
@jow- did you see my comment above? This didn't fix it for me

jow-:
Merged again via 42d3b3d4ca214d967eabb1138be386ddd0665726 with fix for redirects included.

@f00b4r0 - your case still requires an explicit `option family any` to get rid of the `meta nfproto ipv4` match. Implicit IPv4-only redirects are for backwards compat with fw3 which did not support IPv6 NAT.

f00b4r0:
@jow- will test fix soon but IMHO it feels odd to carry on that fw3 IPv4-only legacy forever. I don't know how this will be handled in Luci, but when defining a configuration in plain uci I think the basic expectation is that unless explicitly restricted (using `option family`) the redirect applies to both v4 and v6. It also means that `redirect` and `rule` will have different default behaviours on this which is yet another source of confusion.

Idea: for automatic upgrades from fw3 to fw4, you could have a global "disable ipv6 nat" switch that's set in uci-defaults when upgrading.
HTH

f00b4r0:
To clarify [jow's comment](https://github.com/openwrt/firewall4/pull/35#issuecomment-2730092217) for anyone else reading this: this works if `option family 'any'` is specified for *both* the `ipset` and `redirect` configs. HTH

PR #36, opened 2024-09-02
[scripts] Remove unnecessary logic from stop script

brada4:
Always remove table and service state to have consistent cleaned up system state on exit even if user intentionally dropped our table. Removes dependency on grep as a consequence.

Ref: https://wiki.nftables.org/wiki-nftables/index.php/Configuring_tables
Supersedes: https://github.com/openwrt/firewall4/pull/33

### DIFF
--- a/root/sbin/fw4
+++ b/root/sbin/fw4
@@ -49,12 +49,9 @@ stop() {
 	{
 		flock -x 1000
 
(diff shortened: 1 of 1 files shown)


brada4:
Hi @jow- 
According to "upstream documentation" no need to perform any checks before deleting table since kernel 4.9.0.

PR #37, opened 2024-09-10
Limit mss fixup to connection start

brada4:
Use conntrack flow packet counter to limit mss fixup filter to the very start of connection flows where adjustable syn/syn can possibly appear.

Two initial packets +
> sysctl net.ipv4.tcp_syn_retries net.ipv4.tcp_synack_retries
net.ipv4.tcp_syn_retries = 6
net.ipv4.tcp_synack_retries = 5

### DIFF
--- a/root/usr/share/firewall4/templates/zone-mssfix.uc
+++ b/root/usr/share/firewall4/templates/zone-mssfix.uc
@@ -1,3 +1,4 @@
+ct packets < 14 {%+ -%}
 {%+ if (rule.family): -%}
 	meta nfproto {{ fw4.nfproto(rule.family) }} {%+ endif -%}
 {%+ include("zone-match.uc", { egress, rule }) -%}
--- a/tests/01_configuration/01_ruleset
+++ b/tests/01_configuration/01_ruleset
@@ -269,7 +269,7 @@ table inet fw4 {
 
 	chain mangle_postrouting {
 		type filter hook postrouting priority mangle; policy accept;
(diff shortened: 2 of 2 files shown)


brada4:
Other heuristics take twice bytecode like !assured/assured & ct tcp, Rises throughput ~10% in default case both unicast and forward.

whiskerp:
Any idea when your fw4 improvements will be accepted into a build with the
new build 23.05.5 having no firewall updates? They are still all pending
incorporation into master by jow.

Thanks
Peter

brada4:
He comes here twice a year, if he misses few watch shifts ill make a branch and add readme ;-)

jow-:
Shouldn't the hardcoded `14` value be calculated dynamically if it relates to `net.ipv4.tcp_syn_retries` and `net.ipv4.tcp_synack_retries` ?

Also this looks like very specific, hard to maintain low level optimizations - what's the significance of this change? Are there any throughput tests?

brada4:
Will adapt in a week

brada4:
This is about state of internet. Windows limits all retransmits to 5  to https://support.microsoft.com/en-us/topic/how-to-modify-the-tcp-ip-maximum-retransmission-time-out-7ae0982a-4963-fa7e-ee79-ff6d0da73db8
Thus one obtained from linux covers it.
If you achieve more than few retransmits under normal operation, mss size is least of your woes.

brada4:
This can be done even quicker, no point keeping it another yesr

PR #38, opened 2024-10-06
Do not explicitly accept unsolicited echo-reply6

brada4:
ICMPv6 echo-reply is already handled by conntrack after first echo-request.


brada4:
firewall3 idem

PR #49, opened 2025-02-16
ruleset: do not reject invalid (out of state) packets

### LINKED ISSUE #13340: Firewall: Significang packet loss on WAN when using default rule "wan Forward reject", solution: "wan Forward drop"
### Describe the bug

I hat issues with significant WAN package loss on my setup. It ranged from 10% to 40% pings not getting a reply.

First of all I found the solution/workaround to this:
**Switching the default wan forward rule from reject to drop solves all my issues!**

wan forward drop

I also could stop the firewall under **System > Startup** and hat no more package loss when pinging.

So far I could exclude issues with
DCHP client v4 or v6 (Tested WAN on static IP)
my Cable Modem (Tested two different models)
the physical eth ports (Tested all 4 ports for wan)

My Setup:
Chinese Intel n5105 NUC with 4x I225-V (rev 03) as a Proxmox hypervisor

The OpenWrt VM network setup:
LAN = virtio bridge to the first I255-V
WAN = PCI pass-thru to a second I225-V

I also tested a setup with only PCI pass-thru, same issue.

I also tested OpenWrt 23.05.0-rc3/targets/x86/64/generic-squashfs-combined-efi.img.gz, same issue

### OpenWrt version

r20134-5f15225c1e

### OpenWrt target/subtarget

x86/64

### Device

QEMU Standard PC (i440FX + PIIX, 1996)

### Image kind

Official downloaded image

### Steps to reproduce

Using the default image with default settings.

### Actual behaviour

Significant package loss on WAN with the default firewall configuration

### Expected behaviour

No package loss

### Additional info

_No response_

### Diffconfig

_No response_

### Terms

- [X] I am reporting an issue for OpenWrt, not an unsupported fork.

brada4:
We cannot fix invalid (checksum, out of state, short) packets by replying them with valid packet. Lets keep grief to bare minimum.

Part-fixes: https://github.com/openwrt/openwrt/issues/13340
(the "new" valid packets would have been replied with spoofed packet but those should have been relatively rare to make problem less noticeable)

Changes jump to goto - there is no "next" rule to reach after connection is reset and forgotten.

### DIFF
--- a/root/usr/share/firewall4/templates/ruleset.uc
+++ b/root/usr/share/firewall4/templates/ruleset.uc
@@ -126,7 +126,7 @@ table inet fw4 {
 		{%+ include("zone-jump.uc", { fw4, zone, rule, direction: "input" }) %}
 {% endfor; endfor %}
 {% if (fw4.input_policy() == "reject"): %}
-		jump handle_reject
+		goto handle_reject
 {% endif %}
 {% fw4.includes('chain-append', 'input') %}
 	}
@@ -147,7 +147,7 @@ table inet fw4 {
 {% endfor; endfor %}
 {% fw4.includes('chain-append', 'forward') %}
 {% if (fw4.forward_policy() == "reject"): %}
-		jump handle_reject
+		goto handle_reject
 {% endif %}
 	}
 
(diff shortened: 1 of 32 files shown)


brada4:
@f00b4r0 tell me if i am wrong tyy

brada4:
Alternative - provide this in nftables.d example but as active rule?

Issue #51, opened 2025-02-19
Wrong include order for ruleset-prepend


somerandoname:
According to the [documentation](https://openwrt.org/docs/guide-user/firewall/firewall_configuration#includes_2203_and_later_with_fw4), custom includes with the `ruleset-prepend` option should be placed "before the fw4 table definition". However, this is currently not the case.

This seems to be caused by the rules being included after the fw4 table was initialized, which effectively makes this option behave nearly the same as `ruleset-append`:
```javascript
table inet fw4
flush table inet fw4
{% if (fw4.check_flowtable()): %}
delete flowtable inet fw4 ft
{% endif %}
{% fw4.includes('ruleset-prepend') %}
```

Changing the order fixes this on my system:
```javascript
{% fw4.includes('ruleset-prepend') %}
table inet fw4
flush table inet fw4
{% if (fw4.check_flowtable()): %}
delete flowtable inet fw4 ft
{% endif %}
``` 

https://github.com/openwrt/firewall4/blob/18fc0ead19faf06b8ce7ec5be84957278e942dfa/root/usr/share/firewall4/templates/ruleset.uc#L8C1-L13C38

brada4:
I am out of ideas how actual implementation could look, because unless you make include file flush table it would be repeatedly adding dame chains+rules over and over again.

somerandoname:
@brada4 This is presently already the case. Rulesets outside the `inet fw4` context have to be "manually" flushed. As far as I am aware, the reordering shouldn't change anything in that regard.

brada4:
Mind to make a PR? Actual documentation misses out that table-pre equals /etc/nftables.d/

brada4:
Mail me if you want proper credit.

jow-:
Under which circumstances is the current inclusion order a problem? Can we rule out thatm oving it will alter the behaviour of existing configs?

brada4:
Description clearly says it is before table .. fw4, while at present it is next to table-prepend separated by some initial setup routines.

jow-:
Well `ruleset-prepend` *is* before and *outside*  `table inet fw4 { ... }` while `table-prepend` is *inside* `table inet fw4 { ... }` so a different scope.

So asking again, under which circumstances is a preceding forward declaration and flushing of `table inet fw4` a problem for an include that is solved by moving the include before the forward declaration?

To me it seems as if the proposed change will change the behavior to be less deterministic.

- before: `table inet fw4` is guaranteed to exist but empty
[...]

PR #52, opened 2025-02-20
[ruleset] Relocate ruleset-prepend include before our table

brada4:
Issue described in https://github.com/openwrt/firewall4/issues/51 and fix proposed by @somerandomname

Fixes: https://github.com/openwrt/firewall4/issues/51
Fixes: https://github.com/openwrt/firewall4/commit/11256ff0374fb594e31b0a4e3857f3810ba2933d

### DIFF
--- a/root/usr/share/firewall4/templates/ruleset.uc
+++ b/root/usr/share/firewall4/templates/ruleset.uc
@@ -5,12 +5,12 @@
 	let zones_with_limits = filter(fw4.zones(), z => z.log_limit);
 -%}
 
+{% fw4.includes('ruleset-prepend') %}
 table inet fw4
 flush table inet fw4
 {% if (fw4.check_flowtable()): %}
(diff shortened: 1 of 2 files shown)


somerandoname:
Hi, it looks like there are situations where the `inet fw4` table already exists before this script gets executed. Mainly after device restarts, and possibly hot-plug events.

A solution could be to just outright delete the table before the include:
```
table inet fw4
delete table inet fw4
{% fw4.includes('ruleset-prepend') %}
```

brada4:
Reload supposedly is lighter than restart. Not by much though. 
Intetestingly there is no magic flag to tell reload from restart in custom table that can reorder different inet(and 4 and 6) tables.

jow-:
> A solution could be to just outright delete the table before the include:

Which would essentially undo most of this PR, I'll retain the original behavior which consistently ensures that the fw4 table exists but is empty.

PR #59, opened 2025-04-27
ruleset: drop invalid packets at mangle priority

brada4:
Use earliest hook after conntrack (-200) to drop invalid packets

As a consequence they are not processed by conntrack at all once identified

Add diagnostic counters, hopefully hinting users in direction of net.netfilter.nf_conntrack_log_invalid=255

Obsoletes: 19a8caf614ec338513e58535ea02c6ee52988170
Reverts: 19a8caf614ec338513e58535ea02c6ee52988170

### DIFF
--- a/root/etc/config/firewall
+++ b/root/etc/config/firewall
@@ -3,6 +3,7 @@ config defaults
 	option input		REJECT
 	option output		ACCEPT
 	option forward		REJECT
+	option drop_invalid	1
 # Uncomment this line to disable ipv6 rules
 #	option disable_ipv6	1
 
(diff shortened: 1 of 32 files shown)


brada4:
second 3rd of https://github.com/openwrt/firewall4/pull/22
Blocks guarding offload with ct state, ie last part of those.

Careful rebasing on top of https://github.com/openwrt/firewall4/pull/56 - iif "lo" should follow established, related accept

Ref: https://wiki.nftables.org/wiki-nftables/index.php/Netfilter_hooks

Ideally this could be 4 hooks at points where ct is entered pre/out -200 and left post/in +32k. But hooks cost CPU esp on low end platforms.

brada4:
Ahh yes, loopback excluded to emulate fw3 workings, no harm in users running nmap -O against loopback (which is quite efficient ct invalid generator on its own)

brada4:
Please consider https://github.com/brada4/firewall4/commit/aec0dc5606ad84efb2b31dd7c0b797f6cc513828 superseding pr #22 completely.

brada4:
To add to #103

PR #60, opened 2025-05-07
examples: add nftrace example ruleset

brada4:
Add example rules to enable nftrace and elaborate filter example to avoid overrunning user's terminal using defaults
Helps detecting subtle deviations in filtering eg.
https://github.com/openwrt/openwrt/issues/18387#issuecomment-2781442911

### DIFF
--- a/root/etc/nftables.d/10-custom-filter-chains.nft
+++ b/root/etc/nftables.d/10-custom-filter-chains.nft
@@ -1,7 +1,6 @@
 ## The firewall4 input, forward and output chains are registered with
 ## priority `filter` (0).
 
-
 ## Uncomment the chains below if you want to stage rules *before* the
 ## default firewall input, forward and output chains.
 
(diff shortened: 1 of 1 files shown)


brada4:
Upstream doc: https://wiki.nftables.org/wiki-nftables/index.php/Ruleset_debug/tracing
Maybe better than command example.

root/etc/nftables.d/10-custom-filter-chains.nft
```diff
@@ -37,3 +36,24 @@
...
 # }
 
+## Uncomment the chains below to stage nftrace instrumentation
+## at the beginning of  both base chains filtering packets first
+## Example enables tracing of DNS packets to limit output volume
+## Adjust to Your requirements and observe output using
+## root@OpenWRT:~# nft monitor trace
+
+#     chain raw_prerouting {
+#     type filter hook prerouting priority raw; policy accept;
```
lantis1008, line 46: Should this be `priority raw - 1` so that it catches before the default NoTrack chains that fw4 adds? So the trace starts at the very beginning.

brada4: It would be necessary if one had multiple raw tables. Now I prepend rule to a single raw table we got to re-use existing hook.
3 main base chains change hook default action and require x - 1 additional hook.

PR #61, opened 2025-05-11
scripts: restore fw3 ct flush behavior

brada4:
Flush conntrack if fw4 is started with fw4 table absent
Approximates fw3 ct flush when no iptables rules are present
Prevents (deletes) eternal ghost states created at early boot.
Also treat flush action called stopping the service, emulate 'destroy table' to to stay v23 compatible

### DIFF
--- a/root/sbin/fw4
+++ b/root/sbin/fw4
@@ -60,11 +60,9 @@ flush() {
 	{
 		flock -x 1000
 
-		local dummy family table
-		nft list tables | while read dummy family table; do
-			nft delete table "$family" "$table"
-		done
-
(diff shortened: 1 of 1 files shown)


brada4:
Raised here first
https://forum.openwrt.org/t/firewall-control-over-established-sessions/228684
~~Request to fully document `nft table create` upstream
https://bugzilla.netfilter.org/show_bug.cgi?id=1800
Alternative would be to create and add test chain which fails in absence of parent table~~

brada4:
Changed to draft, got better (performant) idea in works not needeing `echo f` kernel patch.

PR #62, opened 2025-05-15
config: Make DHCP and DHCPv6 response rules strict

brada4:
Non-server DHCPx responses indicated by non-standard sport are already discarded by client, reflect that in firewall rule avoiding unnecessary ct state buildup wasting ct resources

### DIFF
--- a/root/etc/config/firewall
+++ b/root/etc/config/firewall
@@ -33,8 +33,17 @@ config rule
 	option name		Allow-DHCP-Renew
 	option src		wan
 	option proto		udp
+	option src_port		67
 	option dest_port	68
 	option target		ACCEPT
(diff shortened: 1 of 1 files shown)


root/etc/config/firewall
```diff
@@ -59,10 +68,19 @@ config rule
 	option name		Allow-DHCPv6
 	option src		wan
 	option proto		udp
+	option src_port		547
 	option dest_port	546
 	option family		ipv6
 	option target		ACCEPT
 
+config ruke
```
feckert:
```suggestion
config rule
```

brada4: amended with typo corrected, thanks for spotting.

PR #63, opened 2025-05-23
config: mark unsupported fw3 options correctly

brada4:
- mark sysctl-based fw3 options UNSUPPORTED, emitting warning
- remove notion of unsupported option from default config file

No change in ruleset logic.


brada4:
Related - restore lost options
https://github.com/openwrt/openwrt/pull/18895

PR #64, opened 2025-05-25
ruleset: skip _iifname match for !tcp in mss fix

brada4:
Add explicit l4proto match before _ifname to avoid burning cycles for other protocols, eliminating measurable (iperf3) udp re-ordering Displayed back rules show pessimal combo even new one is loaded.

Fixes: couple of issues pertaining UDP ordering in games and cheap loudspeakers TBA

### DIFF
--- a/root/usr/share/firewall4/templates/zone-mssfix.uc
+++ b/root/usr/share/firewall4/templates/zone-mssfix.uc
@@ -1,5 +1,6 @@
 {%+ if (rule.family): -%}
 	meta nfproto {{ fw4.nfproto(rule.family) }} {%+ endif -%}
+	meta l4proto tcp {%+ -%}
 {%+ include("zone-match.uc", { egress, rule }) -%}
 tcp flags syn / syn,fin,rst tcp option maxseg size set rt mtu {%+ if (zone.log & 2): -%}
 	log prefix "MSSFIX {{ zone.name }} out: " {%+ endif -%}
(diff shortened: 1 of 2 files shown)


brada4:
in
```
table inet t {
chain c {
oifname "eth1" tcp flags & (fin | syn | rst) == syn tcp option maxseg size set rt mtu
meta l4proto tcp oifname "eth1" tcp flags & (fin | syn | rst) == syn tcp option maxseg size set rt mtu
}
}
```
out
```
inet t c
  [ meta load oifname => reg 1 ]
  [ cmp eq reg 1 0x31687465 0x00000000 0x00000000 0x00000000 ]
  [ meta load l4proto => reg 1 ]
  [ cmp eq reg 1 0x00000006 ]
...
inet t c
  [ meta load l4proto => reg 1 ]
  [ cmp eq reg 1 0x00000006 ]
  [ meta load oifname => reg 1 ]
  [ cmp eq reg 1 0x31687465 0x00000000 0x00000000 0x00000000 ]
...
```

brada4:
Kind of still sucks for tcp established connections.

Djfe:
do you mean by that that this is just a workaround?

brada4:
tcp mss fix will still do the name comparison for each tcp packet. This just removes useless check for measurably bad offloaded udp streams. just a side observation in the context.

brada4:
Swapped to draft, i got better idea, but this is still minimally intrusive for stable trains

Issue #53, opened 2025-03-02
Firewall is not working for IPSec NAT-T


kouhj:
Tested two OpenWRT 24.10 routers to setup a site to site IPSec connection with Strongswan when the WAN interfaces are over NATed network. The result is the connection can be setup properly, and the following commands all reported normal results:
`swanctl --list-sas`
`ip xfrm state`
`ip xfrm policy`
`ip route show table 220`

**BUT the normal site to site traffic just can get through when tested with ping.**
If the firewall is stopped on the router that initiates the ping, it worked just fine to reach the other router. And unreachable again if the firewall is re-enabled.

Followed these articles,
https://docs.strongswan.org/docs/latest/howtos/forwarding.html
https://docs.strongswan.org/docs/latest/howtos/trafficDumps.html
and found that the LAN traffic can pass the `postrouting` chain of the `mangle` table with this command `iptables -t mangle -I POSTROUTING -m policy --pol ipsec --dir out -j NFLOG --nflog-group 5`. The key seems how to add a nftables rule in `POSTROUTING` chain of the `nat` table like the command `iptables -t nat -I POSTROUTING -m policy --pol ipsec --dir out -j ACCEPT` pointed out in the Strongswan doc mentioned above.

Tried the following and no luck:
`nft insert rule inet fw4 srcnat oif WAN ipsec out reqid 1 counter accept`
`nft insert rule inet fw4 mangle_postrouting oif WAN ipsec out reqid 1 counter accept`

The reqid 1 is confirmed with `swanctl --list-sas`

Any ideas how to make IPSec NAT-T work with fw4? Thanks!

brada4:
Does iptables-translate (or iptables-nft) do the conversion for you?
https://git.netfilter.org/iptables/tree/extensions/libxt_policy.txlate

kouhj:
No. this tools does not work:
`iptables-translate -t mangle -I POSTROUTING -m policy --pol ipsec --dir out -j ACCEPT`
`nft # -t mangle -I POSTROUTING -m policy --pol ipsec --dir out -j ACCEPT`

I did the conversion by myself as in the previous post by adding a rule to `mangle_postrouting ` chain or `srcnat` before the masquerading. The rule can be added but it did not ACCEPT the packet before it SNAT the sender IP, so the original unencapsulated packet was sent to the WAN device. So the IPSec communication failed.

brada4:
Maybe something like `mieta ipsec exists` `meta ipsec missing` ?

[...]

PR #65, opened 2025-05-31
config: strict ip6 ipsec matches

brada4:
ip4 part never served any purpose except confusing users eg https://github.com/openwrt/firewall4/issues/53
Permit only IKE conformant to RFC, ip4 NAT traversal has different port and terminating on the router can be handled by ipsec policy filtering.

Fixes: https://github.com/openwrt/openwrt/commit/8fb39f168249f15697ab930245ad08ea66a1b926

### DIFF
--- a/root/etc/config/firewall
+++ b/root/etc/config/firewall
@@ -113,20 +113,31 @@ config rule
 	option target		ACCEPT
 
 config rule
-	option name		Allow-IPSec-ESP
+	option name		Allow-IPv6-IPSec-ESP
 	option src		wan
 	option dest		lan
 	option proto		esp
+	option family		ipv6
 	option target		ACCEPT
 
 config rule
-	option name		Allow-ISAKMP
+	option name		Allow-IPv6-IKE
+	option src		wan
+	option dest		lan
+	option src_port		500
+	option dest_port	500
+	option proto		udp
+	option family		ipv6
+	option target		ACCEPT
+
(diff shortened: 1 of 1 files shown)


root/etc/config/firewall
```diff
@@ -113,20 +113,31 @@ config rule
...
-	option name		Allow-IPSec-ESP
+	option name		Allow-IPv6-IPSec-ESP
 	option src		wan
 	option dest		lan
 	option proto		esp
+	option family		ipv6
 	option target		ACCEPT
 
 config rule
-	option name		Allow-ISAKMP
```
abajk, line 123: I think this breaks Voice over Wi-Fi. Correct me if I'm wrong.

abajk (changes requested):

root/etc/config/firewall
brada4: Voice over wifi is SIP, not related to ipsec.

abajk: Please read how WiFi Calling works. It is not pure SIP. The traffic is encapsulated in IPSEC tunnel.

brada4: Absolutely unrelated to IPSEC NAT Traversal encapsulating ipsec in 4500/udp.
Examine RFC this comes from https://www.rfc-editor.org/rfc/rfc6092

PR #74, opened 2025-10-02
fw4: enable mac ranges for rule

ckorber:
nft supports handling mac ranges and therefore this commit changes fw4 to support that feature. The src_mac is now allowed to be a address with CIDR notation. If no CIDR notation is configured, the old logic is applied.

So this is now possible:
```
option src_mac '00:11:AA:BB:CC:DD-00:11:AA:FF:FF:FF'
```

Also negation with '!' at the beginning to block every MAC not in the range is possible.

### DIFF
--- a/root/usr/share/ucode/fw4.uc
+++ b/root/usr/share/ucode/fw4.uc
@@ -1188,15 +1188,27 @@ return {
 	},
 
 	parse_mac: function(val) {
-		let mac = this.parse_invert(val);
-		let m = mac ? match(mac.val, /^([0-9a-f]{1,2})[:-]([0-9a-f]{1,2})[:-]([0-9a-f]{1,2})[:-]([0-9a-f]{1,2})[:-]([0-9a-f]{1,2})[:-]([0-9a-f]{1,2})$/i) : null;
+		let macregex = /^([0-9a-f]{1,2})[:-]([0-9a-f]{1,2})[:-]([0-9a-f]{1,2})[:-]([0-9a-f]{1,2})[:-]([0-9a-f]{1,2})[:-]([0-9a-f]{1,2})$/i;
+		let rangeregex = /^([0-9a-f]{1,2})[:-]([0-9a-f]{1,2})[:-]([0-9a-f]{1,2})[:-]([0-9a-f]{1,2})[:-]([0-9a-f]{1,2})[:-]([0-9a-f]{1,2})(\-?)([0-9a-f]{1,2})[:-]([0-9a-f]{1,2})[:-]([0-9a-f]{1,2})[:-]([0-9a-f]{1,2})[:-]([0-9a-f]{1,2})[:-]([0-9a-f]{1,2})?$/i;
 
-		if (!m)
+		let mac = this.parse_invert(val);
+		let m = mac ? match(mac.val, macregex) : null;
+		let mrange = mac ? match(mac.val,rangeregex) : null;
+
+		if (!m && !mrange)
 			return null;
 
-		mac.mac = sprintf('%02x:%02x:%02x:%02x:%02x:%02x',
+		if ( mrange && '-' === mrange[7]) {
+			mac.mac = sprintf('%02x:%02x:%02x:%02x:%02x:%02x-%02x:%02x:%02x:%02x:%02x:%02x',
+		                  hex(mrange[1]), hex(mrange[2]), hex(mrange[3]),
+		                  hex(mrange[4]), hex(mrange[5]), hex(mrange[6]),
+		                  hex(mrange[8]), hex(mrange[9]), hex(mrange[10]),
+		                  hex(mrange[11]), hex(mrange[12]), hex(mrange[13]));
+		} else {
+			mac.mac = sprintf('%02x:%02x:%02x:%02x:%02x:%02x',
 		                  hex(m[1]), hex(m[2]), hex(m[3]),
 		                  hex(m[4]), hex(m[5]), hex(m[6]));
+		}
 
 		return mac;
 	},


brada4:
CIDR is wrong approach for MAC address, you have to be able to extract 2 youngest bits of oldest byte.

ckorber:
So if I understand you correctly you mean that it is not useful to enable the whole range, because the first byte contains important information that should not be manipulated, am I right?

I can modify it to only allow ranges in the last three youngest bytes (so not the oui).
Is that also your intention?

brada4:
The grouping of MAC addresses with mask is very welcome, just that presuming CIDR is wrong. Yo uare on the right path, just that took a bit odd turn.
DOC: https://en.wikipedia.org/wiki/MAC_address#Ranges_of_group_and_locally_administered_addresses

Example `src 00:00:00:00:00:00/01:00:00:00:00:00 drop` would drop unrealistic packets sourcing from multicast addresses.

ckorber:
I get what you mean, I think.
The first byte holds valuable information that shouldn't be touched.

But if I limit it to bytes after the oui, we should be safe, right?

brada4:
Just use full mask, thats all I wanted to tell. There is no CIDR for mac addresses.

brada4:
Something wrong with parentheses ;-)

ckorber:
Better? :)

I dropped the CIDR approach and let it parse for ranges that are set or return as usual.
If the range is wrong (that is if the lower is on the right) it returns nothing.

~~EDIT: Oh I didn't push again.~~

brada4:
You can compile ucode and ./run_tests.sh (it is a bit tricky first time, you have to disable all or most extensions via cmake-gui or parameters)

brada4:
You dont have to check special case with 0 or -1 mask that makes no-op-ish rules

root/usr/share/ucode/fw4.uc
```diff
@@ -2420,12 +2448,18 @@ return {
...
 				dports_neg: map(filter_neg(dports), this.port),
-				smacs_pos: map(filter_pos(rule.src_mac), m => m.mac),
-				smacs_neg: map(filter_neg(rule.src_mac), m => m.mac),
+				smacs_pos: map(filter_pos(rule.src_mac), this.mac_range),
+				smacs_neg: map(filter_neg(rule.src_mac), this.mac_range),
 				icmp_types: map(icmptypes, i => (family == 4 ? i.type : i.type6)),
 				icmp_codes: map(icmpcodes, ic => `${(family == 4) ? ic.type : ic.type6} . ${(family == 4) ? ic.code_min : ic.code6_min}`)
 			};
 
+			if (!length(r.smacs_pos))
```
ckorber: > You dont have to check special case with 0 or -1 mask that makes no-op-ish rules

Are you talking about these lines?

EDIT: The `./run_tests.sh` was fine. All tests passed.

ckorber: If I understand you correctly now, you meant that I don't have to check for correctness because fw4 already handles that, am I right?

ckorber:
Does that work for you @brada4 ?

brada4:
Wait for @jow- to examine it.

ckorber:
ping @jow-

systemcrash:
Advise you reword commit message:
`Also negation with '!' at the beginning to block every MAC not in the range is possible.`
to
`Also negation with '!' at the beginning to block every MAC in the range is possible.`

ckorber:
But I meant what I wrote?
AFAIK the '!' negates the range and applies to the opposite, so it handles every MAC not in the range.

I could change it to:
`Also negation with '!' at the beginning to handle every MAC not in the range is possible.`

brada4:
It is double no in your sentence meaning YES in some languages...

ckorber:
```
If a range is specified with '!', like '!00:11:AA:00:00:00-00:11:AA:FF:FF:FF'
it handles every MAC that is not within the range.
```
I rephrased it. Is it now clearer?

systemcrash:
Handle could be ambiguous depending on the action requested. The initial wording of ! == block (within the range) is clear and concrete. Less cognitive load.

brada4:
exclamation mark is called negation  in iptables doc 
eg "negative matching using exclamation mark is supported"

ckorber:
@brada4 : I used your suggestion for the commit message.

Any other way to progress this PR? I think @jow- is not available at the moment.

systemcrash:
One potential pitfall with ambiguity: it becomes difficult to differentiate between MAC and hyphen separator when the MACs themselves use hyphen - between octets. :)

`00-11-AA-00-00-00-00-11-AA-FF-FF-FF`. 

Although I can't think of a better separator. Hyphen is typically used to denote ranges. If this is deemed problematic, how about underscore? No, that's a binder. Tilde?

`00-11-AA-00-00-00~00-11-AA-FF-FF-FF`.

Thoughts @ckorber ?

brada4:
nftables takes all imaginable formats
in
```c
table inet t {
 chain c {
 ether saddr & 01:00:00:00:00:00 == 0x0
  ether saddr "02-00-00-00-00-00"-"04.ff.ff.ff.ff.ff"
 }
}
```
out
```c
table inet t {
        chain c {
                ether saddr & 01:00:00:00:00:00 == 00:00:00:00:00:00
                ether saddr 02:00:00:00:00:00-04:ff:ff:ff:ff:ff
        }
}
```
0x0 in first line overflows if exceeding uint32 so it is not really useful for anything. standard format is with colon, other formats needs to be escaped.

ckorber:
Hate that I didn't see that :smile:
It wouldn't be a problem in luci (as it is with my depending PR openwrt/luci#8017), because it relies on `:` as separator.

Maybe `00-01-AA-00-00-00..00-01-AA-FF-FF-FF` gives the right idea?
But I don't know if this is the best approach?
Maybe changing this adds an additional cognitive load to remember that ranges are noted with `..` as range symbol in fw4.
I think `-` is more intuitive.

Additionally after parsing it transforms to `00:01:AA:00:00:00-00:01:AA:FF:FF:FF` again in fw4.

brada4:
99% of users will supply valid values but we dont want others to break ruleset.

systemcrash:
> I think `-` is more intuitive.

Agreed. Just happier that we at least discussed the potential problem here.

ckorber:
Don't want to push it, but there is no progress on this PR. Can someone review and merge it?
Or is it only possible by @jow-?

systemcrash:
Any maintainer can.

systemcrash:
@ckorber Recommend tweaking the commit message a bit to e.g.

So this is now possible:
```
option src_mac '00:11:AA:00:00:00-00:11:AA:FF:FF:FF'
```
In addition to the original:
```
option src_mac '00:11:AA:00:00:00'
```

ckorber:
I added that. Thanks!
> @ckorber Recommend tweaking the commit message a bit to e.g.
> 
> So this is now possible:
> ```
> option src_mac '00:11:AA:00:00:00-00:11:AA:FF:FF:FF'
> ```
> In addition to the original:
> ```
> option src_mac '00:11:AA:00:00:00'
> ```
>

ckorber:
@nbd168 or @stintel : Do one of you want to review, because you committed to the project in the past? @jow- seems to be unavailable for review.
Thanks in advance!

systemcrash:
Hi @robimarko could we get this in for 25?

root/usr/share/ucode/fw4.uc
```diff
@@ -1189,14 +1189,22 @@ return {
 
 	parse_mac: function(val) {
 		let mac = this.parse_invert(val);
-		let m = mac ? match(mac.val, /^([0-9a-f]{1,2})[:-]([0-9a-f]{1,2})[:-]([0-9a-f]{1,2})[:-]([0-9a-f]{1,2})[:-]([0-9a-f]{1,2})[:-]([0-9a-f]{1,2})$/i) : null;
+		let m = mac ? match(mac.val, /^([0-9a-f]{1,2})[:-]([0-9a-f]{1,2})[:-]([0-9a-f]{1,2})[:-]([0-9a-f]{1,2})[:-]([0-9a-f]{1,2})[:-]([0-9a-f]{1,2})(\-)?([0-9a-f]{1,2})[:-]([0-9a-f]{1,2})[:-]([0-9a-f]{1,2})[:-]([0-9a-f]{1,2})[:-]([0-9a-f]{1,2})[:-]([0-9a-f]{1,2}))?$/i) : null;
```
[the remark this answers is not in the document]
brada4: I mean
```diff
- ^([0-9a-f]{1,2})[:-]([0-9a-f]{1,2})[:-]([0-9a-f]{1,2})[:-]([0-9a-f]{1,2})[:-]([0-9a-f]{1,2})[:-]([0-9a-f]{1,2})$
+ ^([0-9a-f]{1,2}){5}([:-]([0-9a-f]{1,2})$
```

ckorber:
I updated the PR because as I tested it, it no longer supported inserting only one MAC.
Now this is possible again. I split the two cases into two regexes.

ckorber:
It has been a while. Anyone open for review or merging? @systemcrash @robimarko

brada4:
It is @jow-

systemcrash:
I see no reason we cannot take advantage of fw4 capability. paging @robimarko :)

ckorber:
> It is @jow-

On Oct, 25:
> ping @jow-

I got no response from @jow- , so I try to contact other maintainers. :smile:

brada4:
Roughly 3 years without major functional improvements, you can ping whomever you want.

domingues:
> Roughly 3 years without major functional improvements, you can ping whomever you want.

Same lack of response on the usteer repo.

PR #76, opened 2025-10-13
ruleset: add prerouting chain includes

brada4:
Add missing includes around prerouting chain
Used by eg transparent proxies.
Spotted by @reinerotto
Add another missing in raw_prerouting

### DIFF
--- a/root/usr/share/firewall4/templates/ruleset.uc
+++ b/root/usr/share/firewall4/templates/ruleset.uc
@@ -181,6 +181,7 @@ table inet fw4 {
 
 	chain prerouting {
 		type filter hook prerouting priority filter; policy accept;
+{% fw4.includes('chain-prepend', 'prerouting') %}
 {% for (let zone in fw4.zones()): %}
 {%  if (zone.dflags.helper): %}
 {%   for (let rule in zone.match_rules): %}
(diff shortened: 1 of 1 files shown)


em-may:
I could really use this for increasing the TTL of SSDP packets. Currently I'm using `ruleset-post`, but that's not ideal. If this could get merged in and updated in the repo that'd be great.

brada4:
You can add includes in /etc/nftables.d/ with +/-1 priority, that is saved over sysupgrade. There is even example in that directory

em-may:
I already have some in `/usr/share/nftables.d/{ruleset-pre,table-pre,chain-pre/forward,ruleset-post}/`, but as far as I can tell I can't add anything in `/usr/share/nftables.d/chain-post/prerouting` since there are no includes for that chain, thus this PR, right? So for the `prerouting` rules I need, I end up putting them in `/usr/share/nftables.d/ruleset-post` like
```
add rule inet fw4 prerouting <rule>
```
If I'm missing something, please lmk, thanks!

brada4:
Too abstract to say yes or no.

brada4:
something like
```
chain abcd {
 type filter ; hook prerouting priority raw + 1;
 //ttl hl set
}
```
those includes are inside table inet fw4{}, ie dont need table definition.

PR #77, opened 2025-10-17
ruleset: avoid no-op port translation in dnat

brada4:
Correct rport null propagation moving compensatory conditional to reflection snat rule only.
Emit shorthand rules if user did not specify redirect target port
This saves 2 bytecodes in rule and kernel does not do no-op port rewrite.
Also tweak tests to cover changed rules. 
V2 add test case from WIKI https://openwrt.org/docs/guide-user/firewall/fw3_configurations/intercept_dns#dns_redirection
V3 omit unchanged dport in config example

### DIFF
--- a/root/etc/config/firewall
+++ b/root/etc/config/firewall
@@ -155,7 +155,6 @@ config rule
 #	option src_dport	80
 #	option dest			lan
 #	option dest_ip		192.168.16.235
-#	option dest_port	80
 #	option proto		tcp
 
 # port redirect of remapped ssh port (22001) on wan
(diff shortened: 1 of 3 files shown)


brada4:
```
# old
meta nfproto ipv4 tcp dport 53 counter dnat 192.0.2.0:53
# new
meta nfproto ipv4 tcp dport 53 counter dnat 192.0.2.0
```
It could be even shorter with no place for counter abusing nft-nat-s builtin nfproto filter.
```
tcp dport 53 dnat ip 192.0.2.0
```

brada4:
Also can zap remote port if equal and optimize user's rule.

PR #78, opened 2025-10-21
fw4:Allow masquerade to set port

HSXX:
Allow masquerade to set port


brada4:
Please add a test rule + output in https://github.com/openwrt/firewall4/blob/6e33fce6014ed629ac55efd3d8e3c458519f30fa/tests/03_rules/07_redirect#L323 to traverse new code path.

brada4:
Actually looks good in final form...

Neustradamus:
Replaced by:
https://github.com/openwrt/firewall4/pull/79

PR #80, opened 2025-10-21
fw4: Add ICMP protocal for SNAT.

Nicole-Norris:
Added missing code to support ICMP for SNAT.


Nicole-Norris:
ping @jow-

jow-:
Please explain why this is required and show uci config as well as generated nftables rules before/after this change.

brada4:
@jow- to me it looks WiP , one-liner with a test missing in essence.
For a bigger picture on where it could be useful as eg cheap mwan-1 https://github.com/vyos/vyos-1x/blob/current/data/templates/firewall/nftables-cgnat.j2

PR #81, opened 2025-10-27
config: add dest addr restrictions for DHCPv6 rules

AndyChiang888:
Some ISPs may use a GUA or other non-LLA as the source addr for the DHCPv6 response, but the destination addr is always LLA (fe80::/10).
Therefore, adding a dest addr restriction improves security.
See https://forum.mikrotik.com/t/xfinity-comcast-dhcpv6-configuration-change/156031/10


brada4:
Just cross-referncing with other restriction bc changing same lines. https://github.com/openwrt/firewall4/pull/62

AndyChiang888:
For DHCPv6, just limiting the dest addr to a LLA is sufficient to ensure security and compatibility.

brada4:
Mine  is read directly from RFC, but yours indeed is more precise.

brada4:
dhcp clients discard otherbsource ports leaving dangling ct unreplied state for them, so both complement eachother

brada4 (approved):

AndyChiang888:
firewall3 is complete (https://github.com/openwrt/openwrt/commit/4ad22d03429d45f9f5769af58c4521b3ff26815a)
now only firewall4 needs to be merged.

AndyChiang888:
@jow- @nbd168 PTAL

geldot:
Change tested and works fine here.

PR #83, opened 2025-11-12
20-firewall: Execute has_zone and and fw4 in background to avoid slow execution of hotplug.d/iface scripts

### LINKED ISSUE #20723: Very slow ifup performance
### Describe the bug

ifup is extremely slow (taking a couple of seconds per interface).

### OpenWrt version

r28959-29397011cc

### OpenWrt release

24.10.4

### OpenWrt target/subtarget

ramips/mt7621

### Device

Netgear R6260

### Image kind

Official downloaded image

### Steps to reproduce

_No response_

### Actual behaviour

_No response_

### Expected behaviour

_No response_

### Additional info

_No response_

### Diffconfig

```text

```

### Terms

- [x] I am reporting an issue for OpenWrt, not an unsupported fork.

rdevshp:
This pull request fixes https://github.com/openwrt/openwrt/issues/20723


brada4:
Something similar by @ptpt52 (pre-dating my impromptu re-invention)  ? https://github.com/x-wrt/x-wrt/blob/master/package/network/config/firewall4/patches/200-fw4-hotplug-fork.patch

brada4 (changes requested):
You have to limit forking of new copies of reload attempts , rapid flipping of link externally would unnecesarily accumulate waiting background tasks.

rdevshp:
What do you think would be the ideal approach here to avoid excessive reload attempts & unnecessarily accumulate background tasks? 
I am planning to create a new service that actually executes `fw4 -q reload`, and the 20-firewall script would merely send a fw4 reload notification to the new service without actually doing the reload itself. Do you think this is a good idea?

brada4:
It is not about clicks in the interface that go in sync with :rofl: background :rofl: activity, it is about rapidly (I tried 10hz) flipping connected ethernet with zone like wan and building up said activities.
(but totally agree massive cpu time hogs should be out of hotplug)

PR #85, opened 2026-01-07
templates: allow inverted limits

saldry:
Translate inverted limits to nft `limit rate over` which specifies that the rule is matching packets over the rate limit.


brada4:
Why dont you add a single bool parameter that limit is in reverse? Same lines already touched...

saldry:
> Why dont you add a single bool parameter that limit is in reverse? Same lines already touched...

I didn't realize that approach. I'll try it later...

saldry:
Updated. Please check :) @brada4

brada4:
You need to load new parameter in `root/usr/share/ucode/fw4.uc`
Maybe call it limit_over to not invent new nomenclature?

saldry:
> You need to load new parameter in `root/usr/share/ucode/fw4.uc`

https://github.com/openwrt/firewall4/blob/b6e5157527d361f99ad52eaa6da273cb0f2dfd59/root/usr/share/ucode/fw4.uc#L1302-L1330

I'm afraid I don't quite understand what you mean. `limit.invert`/`rv.invert` is available since `parse_limit` and `parse_invert`. And it works well in my tests.

PR #91, opened 2026-03-27
firewall: skip reload on DHCP lease renewal without address change

anand16158:
firewall: skip reload on DHCP lease renewal without address change

Avoid unnecessary firewall reload when only DHCP lease time changes
and interface addresses remain unchanged.

Signed-off-by: Anand Kumar anandvtu16158@gmail.com

### DIFF
--- a/root/etc/hotplug.d/iface/20-firewall
+++ b/root/etc/hotplug.d/iface/20-firewall
@@ -7,7 +7,7 @@ has_zone() {
 }
 
 [ "$ACTION" = ifup -o "$ACTION" = ifupdate ] || exit 0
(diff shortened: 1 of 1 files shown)


brada4:
Good spotting.

brada4:
Please commit --amend ; push --force and fix cm email.

anand16158:
> Please commit --amend ; push --force and fix cm email.

this is done thanks @brada4

geldot:
$IFUPDATE_DATA can reflect more than address change e.g. data supplied dynamically from netifd proto handler e.g. zone supplied by tunnels etc. It isn't safe to skip these

PR #92, opened 2026-04-10
ruleset: avoid port translating local sockets

### LINKED ISSUE #22765: fw4: masquerade incorrectly rewrites source port of router-originated traffic
### Describe the bug

In OpenWrt 25.12, the fw4 masquerade implementation applies NAT source port rewriting to packets **originating from the router itself**, not only to forwarded traffic from LAN clients. This is incorrect behaviour that has the potential to break any service running on the router where the remote peer depends on a specific source port

### OpenWrt version

r32768-b21cfa8f8c

### OpenWrt release

25.12.12

### OpenWrt target/subtarget

x86/64

### Device

Supermicro A2SDi-8C-HLN4F

### Image kind

Self-built image

### Steps to reproduce

1. Configure a WAN zone with `option masq '1'` (the default).
2. Create a kernel L2TP ethernet tunnel (`ip l2tp add tunnel`) bound to a specific UDP port.
3. Bring up the L2TP session and configure an IPv6 address on the tunnel interface.
4. Attempt to communicate through the tunnel from the router itself.

It might be necessary to restart the firewall once or twice with `fw4 restart` to ensure a conntrack table conflict and allocation of a new ephemeral port.

### Actual behaviour

The unguarded masquerade rule (highlighted below using a diff) applies to all IPv4 traffic passing through the `srcnat` chain, including traffic generated by the router's own processes. Despite the source IP already being the WAN interface's assigned address, nftables' `masquerade` processes the packet through the NAT port allocation engine and rewrites the source port to an arbitrary ephemeral value.

  ```diff                                                                                                                                                                                                                                         
   meta nfproto {{ fw4.nfproto(family) }} {%+ if (saddrs && saddrs[0]): -%}
        {{ fw4.ipproto(family) }} saddr {{ fw4.set(map(saddrs[0], fw4.cidr)) }} {%+ endif -%}                                                                                                                                                     
   {%+ if (saddrs && saddrs[1]): -%}          
        {{ fw4.ipproto(family) }} saddr != {{ fw4.set(map(saddrs[1], fw4.cidr)) }} {%+ endif -%}
   {%+ for (let a in (saddrs ? saddrs[2] : [])): -%}                                                                                                                                                                                              
        {{ fw4.ipproto(family) }} saddr & {{ a.mask }} {{ a.invert ? '!=' : '==' }} {{ a.addr }} {%+ endfor -%}                                                                                                                                   
   {%+ if (daddrs && daddrs[0]): -%}                                                                                                                                                                                                              
        {{ fw4.ipproto(family) }} daddr {{ fw4.set(map(daddrs[0], fw4.cidr)) }} {%+ endif -%}                                                                                                                                                     
   {%+ if (daddrs && daddrs[1]): -%}      
        {{ fw4.ipproto(family) }} daddr != {{ fw4.set(map(daddrs[1], fw4.cidr)) }} {%+ endif -%}                                                                                                                                                  
   {%+ for (let a in (daddrs ? daddrs[2] : [])): -%}                                                                                                                                                                                              
        {{ fw4.ipproto(family) }} daddr & {{ a.mask }} {{ a.invert ? '!=' : '==' }} {{ a.addr }} {%+ endfor -%}                                                                                                                                   
  -masquerade comment "!fw4: Masquerade {{ fw4.nfproto(family, true) }} {{ zone.name }} traffic"                                                                                                                                                  
  ```      


### Expected behaviour

`masquerade` should only apply to forwarded traffic from LAN clients whose source IP addresses require translation to the router's public WAN IP. Traffic originating from the router itself already has the correct source IP and **must not** have its source port rewritten.

### Fix

Add `fib saddr type != local` before `masquerade` in `zone-masq.uc`. The `fib` expression performs a kernel FIB lookup on the packet's source address. If the address is locally assigned to the router (type `local`), masquerade is skipped. If it is not, masquerade is applied as normal.

### Patch (`usr/share/firewall4/templates/zone-masq.uc`)

```diff
--- a/templates/zone-masq.uc
+++ b/templates/zone-masq.uc
@@ -10,4 +10,4 @@ meta nfproto {{ fw4.nfproto(family) }} {
  	{{ fw4.ipproto(family) }} daddr != {{ fw4.set(map(daddrs[1], fw4.cidr)) }} {%+ endif -%}
  {%+ for (let a in (daddrs ? daddrs[2] : [])): -%}
  	{{ fw4.ipproto(family) }} daddr & {{ a.mask }} {{ a.invert ? '!=' : '==' }} {{ a.addr }} {%+ endfor -%}
-masquerade comment "!fw4: Masquerade {{ fw4.nfproto(family, true) }} {{ zone.name }} traffic"
+fib saddr type != local masquerade comment "!fw4: Masquerade {{ fw4.nfproto(family, true) }} {{ zone.name }} traffic"
```

### Additional info

### Debug output - IP addresses, MAC addresses and domain names sanitised for privacy reasons

I suspect that the consistent port remapping occurred because after fw4 restarts (which flushes the table and recreate rules), new conntrack entries were being created for the L2TP flow and the NAT engine was allocating a fresh ephemeral port rather than preserving 31392

### Tunnel configuration

```
root@openwrt:~# ip l2tp show tunnel
Tunnel 5, encap UDP
  From A.A.A.A to B.B.B.B
  Peer tunnel 5
  UDP source / dest ports: 31392/31392
  UDP checksum: disabled

root@openwrt:~# ip l2tp show session
Session 1000 in tunnel 5
  Peer session 1000, tunnel 5
  interface name: l2tpeth0
  offset 0, peer offset 0
```

### Interface configuration (`/etc/config/network`)

```
config interface 'wan6'
    option proto 'static'
    option mtu '1420'
    list ip6addr 'fd3e:7a1b:2c9d:4e5f::1/127'
    option ip6gw 'fd3e:7a1b:2c9d:4e5f::'
    option ip6prefix '2001:db8:9:66::/64'
    option device 'l2tpeth0'
```

### Interface information

```
root@openwrt:~# ip addr show dev l2tpeth0
116: l2tpeth0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1420 qdisc fq_codel state UNKNOWN group default qlen 1000
    link/ether c4:de:f1:23:45:67 brd ff:ff:ff:ff:ff:ff
    inet6 fd3e:7a1b:2c9d:4e5f::1/127 scope global
       valid_lft forever preferred_lft forever
    inet6 fe80::c6de:f1ff:fe23:4567/64 scope link proto kernel_ll
       valid_lft forever preferred_lft forever
```

### Ping from router to remote tunnel endpoint — 100% packet loss

```
root@openwrt:~# ping6 -c2 fd3e:7a1b:2c9d:4e5f::
PING fd3e:7a1b:2c9d:4e5f:: (fd3e:7a1b:2c9d:4e5f::): 56 data bytes

--- fd3e:7a1b:2c9d:4e5f:: ping statistics ---
2 packets transmitted, 0 packets received, 100% packet loss
```

### tcpdump on WAN interface showing source port rewritten by masquerade

The L2TP socket is bound to port 31392, but masquerade rewrites the source port to ephemeral port 23525. The remote peer rejects these packets as the source port does not match the established tunnel.

```
root@openwrt:~# tcpdump -i pppoe-wan udp and port 31392
listening on pppoe-wan, link-type LINUX_SLL (Linux cooked v1), snapshot length 262144 bytes
08:14:58.094176 IP A.A.A.A.my.isp.net.23525 > my.server.net.31392: UDP, length 98
08:14:59.114180 IP A.A.A.A.my.isp.net.23525 > my.server.net.31392: UDP, length 98
```

### Ping from LAN IPv6 device — 100% packet loss

```
user@landevice:~$ ping -6 2600::
PING 2600::(2600::) 56 data bytes
^C
--- 2600:: ping statistics ---
3 packets transmitted, 0 received, 100% packet loss, time 2034ms
```

---

### Workaround applied: prevent source port rewriting for tunnel traffic and flush conntrack

```
root@openwrt:~# nft insert rule inet fw4 srcnat_wan meta nfproto ipv4 udp sport 31392 return
root@openwrt:~# echo f > /proc/net/nf_conntrack
```

### Ping from router to remote tunnel endpoint — succeeding after workaround

```
root@openwrt:~# ping6 fd3e:7a1b:2c9d:4e5f::
PING fd3e:7a1b:2c9d:4e5f:: (fd3e:7a1b:2c9d:4e5f::): 56 data bytes
64 bytes from fd3e:7a1b:2c9d:4e5f::: seq=788 ttl=64 time=4.005 ms
64 bytes from fd3e:7a1b:2c9d:4e5f::: seq=789 ttl=64 time=2.890 ms
64 bytes from fd3e:7a1b:2c9d:4e5f::: seq=790 ttl=64 time=3.451 ms
64 bytes from fd3e:7a1b:2c9d:4e5f::: seq=791 ttl=64 time=2.804 ms
64 bytes from fd3e:7a1b:2c9d:4e5f::: seq=792 ttl=64 time=3.505 ms
64 bytes from fd3e:7a1b:2c9d:4e5f::: seq=793 ttl=64 time=2.618 ms
64 bytes from fd3e:7a1b:2c9d:4e5f::: seq=794 ttl=64 time=3.292 ms
64 bytes from fd3e:7a1b:2c9d:4e5f::: seq=795 ttl=64 time=3.859 ms
```

### tcpdump after workaround — bidirectional traffic on correct port

```
root@openwrt:~# tcpdump -i pppoe-wan udp port 31392
listening on pppoe-wan, link-type LINUX_SLL (Linux cooked v1), snapshot length 262144 bytes
08:15:38.898702 IP my.server.net.31392 > A.A.A.A.my.isp.net.31392: UDP, length 98
08:15:39.054221 IP A.A.A.A.my.isp.net.31392 > my.server.net.31392: UDP, length 98
08:15:39.901881 IP my.server.net.31392 > A.A.A.A.my.isp.net.31392: UDP, length 98
08:15:40.074728 IP A.A.A.A.my.isp.net.31392 > my.server.net.31392: UDP, length 98
```

### Ping from LAN IPv6 device — succeeding after workaround

```
user@landevice:~$ ping -6 2600::
PING 2600::(2600::) 56 data bytes
64 bytes from 2600::: icmp_seq=417 ttl=63 time=3.71 ms
64 bytes from 2600::: icmp_seq=710 ttl=63 time=3.76 ms
64 bytes from 2600::: icmp_seq=711 ttl=63 time=3.53 ms
64 bytes from 2600::: icmp_seq=712 ttl=63 time=3.51 ms
64 bytes from 2600::: icmp_seq=713 ttl=63 time=2.99 ms
```

### Diffconfig

```text

```

### Terms

- [x] I am reporting an issue for OpenWrt, not an unsupported fork.

brada4:
avoid allocating extra outgoing ports for local connections

Eliminate unnecesary port translation for locally originated connections
- firstly 2 ports are used for any connection, including single DNS packet
- if you run asterix or dht outgoing connections get randomized and keep ephemeral ports for long after they were relevant. i.e outgoing UDP port is unnecesarily held for 3mins, especially right after priming up dht when it pokes around the nets 1000 pipes at a time.
- once original incoming ct state expires udp based vpns get port-translated and have no chance to ever be kept alive from remote site, and eg normal site2site IKE needs to wait for timeouts to expire to get back to normal
- not a typical usage but dhcp relay with dhcp server behind NAT gets port mangled and has no chance to work
 
fixes: openwrt/openwrt#22765

### DIFF
--- a/root/usr/share/firewall4/templates/ruleset.uc
+++ b/root/usr/share/firewall4/templates/ruleset.uc
@@ -310,6 +310,7 @@ table inet fw4 {
 {% for (let redirect in fw4.redirects("srcnat")): %}
 		{%+ include("redirect.uc", { fw4, zone: null, redirect }) %}
 {% endfor %}
+		iif 0 return
 {% for (let zone in fw4.zones()): %}
 {%  if (zone.dflags.snat): %}
 {%   for (let rule in zone.match_rules): %}
--- a/tests/01_configuration/01_ruleset
+++ b/tests/01_configuration/01_ruleset
@@ -238,6 +238,7 @@ table inet fw4 {
 
 	chain srcnat {
 		type nat hook postrouting priority srcnat; policy accept;
+		iif 0 return
 		oifname "pppoe-wan" jump srcnat_wan comment "!fw4: Handle wan IPv4/IPv6 srcnat traffic"
 	}
 
--- a/tests/01_configuration/02_rule_order
+++ b/tests/01_configuration/02_rule_order
@@ -183,6 +183,7 @@ table inet fw4 {
 
 	chain srcnat {
 		type nat hook postrouting priority srcnat; policy accept;
+		iif 0 return
 	}
 
 
--- a/tests/02_zones/01_policies
+++ b/tests/02_zones/01_policies
@@ -213,6 +213,7 @@ table inet fw4 {
 
 	chain srcnat {
 		type nat hook postrouting priority srcnat; policy accept;
+		iif 0 return
 	}
 
 
--- a/tests/02_zones/02_masq
+++ b/tests/02_zones/02_masq
@@ -206,6 +206,7 @@ table inet fw4 {
 
 	chain srcnat {
 		type nat hook postrouting priority srcnat; policy accept;
+		iif 0 return
 		oifname "zone1" jump srcnat_test1 comment "!fw4: Handle test1 IPv4/IPv6 srcnat traffic"
 		oifname "zone2" jump srcnat_test2 comment "!fw4: Handle test2 IPv4/IPv6 srcnat traffic"
 		oifname "zone3" jump srcnat_test3 comment "!fw4: Handle test3 IPv4/IPv6 srcnat traffic"
--- a/tests/02_zones/03_masq_src_dest_restrictions
+++ b/tests/02_zones/03_masq_src_dest_restrictions
@@ -206,6 +206,7 @@ table inet fw4 {
 
 	chain srcnat {
 		type nat hook postrouting priority srcnat; policy accept;
+		iif 0 return
 		oifname "zone1" jump srcnat_test1 comment "!fw4: Handle test1 IPv4/IPv6 srcnat traffic"
 		oifname "zone2" jump srcnat_test2 comment "!fw4: Handle test2 IPv4/IPv6 srcnat traffic"
 	}
--- a/tests/02_zones/04_masq_allow_invalid
+++ b/tests/02_zones/04_masq_allow_invalid
@@ -131,6 +131,7 @@ table inet fw4 {
 
 	chain srcnat {
 		type nat hook postrouting priority srcnat; policy accept;
+		iif 0 return
 		oifname "zone1" jump srcnat_test1 comment "!fw4: Handle test1 IPv4/IPv6 srcnat traffic"
 	}
 
--- a/tests/02_zones/04_wildcard_devices
+++ b/tests/02_zones/04_wildcard_devices
@@ -317,6 +317,7 @@ table inet fw4 {
 
 	chain srcnat {
 		type nat hook postrouting priority srcnat; policy accept;
+		iif 0 return
 	}
 
 
--- a/tests/02_zones/05_subnet_mask_matches
+++ b/tests/02_zones/05_subnet_mask_matches
@@ -183,6 +183,7 @@ table inet fw4 {
 
 	chain srcnat {
 		type nat hook postrouting priority srcnat; policy accept;
+		iif 0 return
 	}
 
 
--- a/tests/02_zones/06_family_selections
+++ b/tests/02_zones/06_family_selections
@@ -316,6 +316,7 @@ table inet fw4 {
 
 	chain srcnat {
 		type nat hook postrouting priority srcnat; policy accept;
+		iif 0 return
 	}
 
 
--- a/tests/02_zones/07_helpers
+++ b/tests/02_zones/07_helpers
@@ -328,6 +328,7 @@ table inet fw4 {
 
 	chain srcnat {
 		type nat hook postrouting priority srcnat; policy accept;
+		iif 0 return
 		oifname "zone1" jump srcnat_test1 comment "!fw4: Handle test1 IPv4/IPv6 srcnat traffic"
 		oifname "zone2" jump srcnat_test2 comment "!fw4: Handle test2 IPv4/IPv6 srcnat traffic"
 	}
--- a/tests/02_zones/08_log_limit
+++ b/tests/02_zones/08_log_limit
@@ -411,6 +411,7 @@ table inet fw4 {
 
 	chain srcnat {
 		type nat hook postrouting priority srcnat; policy accept;
+		iif 0 return
 		oifname "br-lan" jump srcnat_lan comment "!fw4: Handle lan IPv4/IPv6 srcnat traffic"
 		meta nfproto ipv4 oifname "pppoe-wan" jump srcnat_wan comment "!fw4: Handle wan IPv4 srcnat traffic"
 	}
--- a/tests/03_rules/01_direction
+++ b/tests/03_rules/01_direction
@@ -112,6 +112,7 @@ table inet fw4 {
 
 	chain srcnat {
 		type nat hook postrouting priority srcnat; policy accept;
+		iif 0 return
 	}
 
 
--- a/tests/03_rules/02_enabled
+++ b/tests/03_rules/02_enabled
@@ -107,6 +107,7 @@ table inet fw4 {
 
 	chain srcnat {
 		type nat hook postrouting priority srcnat; policy accept;
+		iif 0 return
 	}
 
 
--- a/tests/03_rules/03_constraints
+++ b/tests/03_rules/03_constraints
@@ -167,6 +167,7 @@ table inet fw4 {
 
 	chain srcnat {
 		type nat hook postrouting priority srcnat; policy accept;
+		iif 0 return
 	}
 
 
--- a/tests/03_rules/04_icmp
+++ b/tests/03_rules/04_icmp
@@ -119,6 +119,7 @@ table inet fw4 {
 
 	chain srcnat {
 		type nat hook postrouting priority srcnat; policy accept;
+		iif 0 return
 	}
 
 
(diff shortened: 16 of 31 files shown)


brada4:
Alternate implementations would not be faster or depend on modules, esp when both meta integer fields are unconditionally populated before conntrack....
```
iif 0 return
  [ meta load iif => reg 1 ]
  [ cmp eq reg 1 0x00000000 ]
  [ immediate reg 0 return ]
meta skuid >= 0 return7
  [ meta load skuid => reg 1 ]
  [ byteorder reg 1 = hton(reg 1, 4, 4) ]
  [ cmp gte reg 1 0x00000000 ]
  [ immediate reg 0 return ]
```
EDIT: upd for v2 patch

stintel:
Neither the change in this PR, nor the proposed unconditional `fib saddr type != local` in openwrt/openwrt#22765 are acceptable as-is, since both change default behavior.

This PR also puts `iif 0 return` in the top-level `srcnat` chain before zone dispatch. That means locally generated packets skip the entire per-zone `srcnat_<zone>` chain, not just the automatically generated zone masquerade rule. In particular, per-zone `config nat` SNAT/masquerade rules and chain-append includes would no longer run for local-origin traffic.

I agree that excluding locally originated traffic from automatic zone masquerading is cleaner, but changing long-standing fw3/fw4 default behavior should not be done lightly, especially in stable releases.

A safer approach would be to add an explicit zone option, e.g. `masq_local`, defaulting to the current behavior for compatibility. Users who need the new behavior can set `option masq_local '0'`, and we can later decide whether new default configs should opt out after a full test cycle.

stintel:
I propose the following change instead: d1fadb5124fd65df57b8e20a4a416220d523fdf3.

I intentionally left out the `meta pkttype multicast return` part. Avoiding masquerade for locally originated packets and skipping multicast traffic are two separate behavioral changes with different rationale and risk. They should be discussed and justified independently, and if needed, handled in separate commits.

The `masq_local` change only addresses the local-origin masquerade case while preserving existing behavior by default and keeping the per-zone `srcnat_<zone>` chains reachable.

brada4:
Masquerade fallback to other port is to save local sockets from random impostor from LAN taking over their data. Was never a race between local sockets themselves to start with.
This ~~major fuckup~~  "long-standing default behavior" should be inverted asap, both users happy with double outgoing port consumption can have their parameter.

stintel:
I’m not claiming the current behaviour is ideal. I’m saying it is not something we should silently change in stable releases.

This goes back to the initial firewall3 import from 2013-02-17, so it has effectively been OpenWrt firewall behaviour for 13 years. fw4 inherited the same model from its initial commit in 2021.

The `masq_local` option gives users affected by this a way to opt out now, without changing behaviour for everyone else. Once this has had some exposure in master, we can discuss changing the default for future releases.

brada4:
Well, come back after you make site to site 500:500 ipv4 ike working using OpenWrt.
No it does not go back that far. It is unnoticed kernel change down the road.

brada4:
Breadcrumb: https://gitlab.com/redhat/centos-stream/src/kernel/centos-stream-9/-/merge_requests/339

explicitly protecting local connections from forwarded connections, not from themselves, happened around last days of fw3 .....

brada4:
Maybe a global setting to enable broken legacy behaviour.

brada4:
Another one bites the dust @stintel 
https://forum.openwrt.org/t/ipsec-tunnel-just-dies-after-some-time-and-cannot-be-brought-back-up-with-ipsec-up/250596

brada4:
Source port preservation  is recommended via RFC 7857, even for miltiple hosts re-using one @stintel please stop pushing your (as in "team") broken crap. Thanks

PR #98, opened 2026-06-08
fw4: support zones and forwardings in procd and netifd firewall data

dangowrt:
fw4 already folds firewall rules published as procd service data and netifd interface data into its ruleset on reload. Let that data also declare zones and the forwardings between them, so a container manager can set up container networking without writing persistent state below /etc/config.

The first three commits fix pre-existing bugs found while validating this: duplicate zone names make nft reject the whole ruleset, unknown spec types vanish silently, and ICMP source NAT ranges (MAP-E portsets) are refused.


blogic:
LGTM

brada4:
Solves root cause in https://github.com/openwrt/firewall4/pull/101 superseding it.
Fixing bugs half is OK, no judgement on IPC extensions.

PS btw clanker review is missing, for sake of fairness it should be un-reviewed by them

PR #99, opened 2026-06-15
ruleset: skip redirect to localhost for local-bound traffic

tleydxdy:
Add fib daddr type != local to redirect rules to avoid creating unnecessary conntrack NAT entries when the destination is already a local address.


brada4 (changes requested):

brada4:
Also needs tests/03_rules/07_redirect changed
Would conflict with https://github.com/openwrt/openwrt/pull/20164

brada4 (commented):
idk (24.10.7) it does not even set dnat flag on direct to self packets.

```c++
        chain dxx {
                type filter hook prerouting priority 32768; 
                meta l4proto { tcp, udp } ct original proto-dst 53 iif "br-lan" counter ct status & dnat vmap { 0x0 counter : continue, dnat counter : continue }
        }
```

brada4:
If you are ready for a ride try to parametrize this

https://github.com/openwrt/openwrt/issues/22287#issuecomment-4074009035

brada4 (commented):
Let me pick up AI rhyme ;-)
The change does not do what it seems to do
- there is a default non-configurable behaviour of kernel that equivalent of `fib daddr . iif type eq local` packets are not subject to redirect
- what the patch excludes from redirect/dnat is dns requests directed to wan or other interface address configured on the router.
- source reference would be in some file at `torvalds/linux/net/netfilter/*`

tleydxdy:
maybe I'm not using the right words here, but it is certainly my intent to exclude from dnat when the traffic targets any of the IPs configured on the router.

brada4:
I dont think it is right, at least not by default.For the counter-example -  You are trying to make kids net one day they learn what is an IP address and bypass "intended restrictions" by entering parents DNS numbers in their computers.
IMHO the expectation is that the router thingie does not shatter in pieces in unskilled hands.

- you can limit source IPs in the DNAT fw4 rule yielding redirect nft rule
- failing that you can make a copy of rule (plus duplicate chain header) into `/etc/nftables.d/*.nft` and adjust to your liking, then disable fw4 rule
- or make this a bool parameter for rule, thats about 100x bigger change than (1) or (2)

or from the other side fw4 ruleset is not covering all corners, and you cannot chain dnat/redirect/snat exceptions between rules, like you can do with normal "access list" style rules.

brada4:
Ahh, yes - all conntrack entries are same size, including 4x IP(6) address , simply NAT state means they are different and some rewrite+re-checksumming is done by netfilter.
Thats why I proposed working out timeouts.

PR #101, opened 2026-07-10
ruleset: make more robust against package mistakes

brada4:
Use "redefine" in place of "define" in zone description macros to make firewall resilient against package mistakenly injecting duplicate zone definitions.

eg https://github.com/openwrt/packages/issues/29962


brada4:
Fixed by 22afe51bed7cf42015eb2403a84b571c3c670ad8

